PR #5, opened 2025-07-09
Define specific error types for RPC methods

benbrandt:
This refactors the RPC error handling from a generic `anyhow::Result` to
method-specific error types using `thiserror`. This allows for more
granular error handling and better type safety, enabling clients to
differentiate between various error conditions for each RPC call.


benbrandt:
Todo:
- add some more serialization tests to verify conversion behavior 
- doc strings for errors

PR #8, opened 2025-07-14
Add protocol version negotiation to initialization

benbrandt:


rust/acp.rs
```diff
@@ -66,6 +67,30 @@ impl AgentConnection {
...
+            major: protocol_version.major,
+            minor: Some(protocol_version.minor),
+            patch: Some(protocol_version.patch),
+            pre: protocol_version.pre.clone(),
+        };
+        let response = self.request(InitializeParams { protocol_version }).await?;
+
+        let server_version = &response.protocol_version;
+
+        if version_requirement.matches(server_version) {
```
benbrandt, line 94: How strict do we want to be here? 
Should we just check the "major" version (first non-zero place) and rely on `capabilities` to validate if certain features are available?

benbrandt: Using this for now as we aren't even to minor versions yet, so this won't matter until then

PR #9, opened 2025-07-14
Add JSONRPC 2.0 version to response

benbrandt:


rust/acp.rs
```diff
@@ -131,8 +131,16 @@ where
...
+#[derive(Debug, Deserialize, Serialize)]
+enum JsonSchemaVersion {
+    #[serde(rename = "2.0")]
+    V2,
+}
+
 #[derive(Debug, Deserialize)]
 struct IncomingMessage<'a> {
+    #[expect(dead_code, reason = "Here for validation, but not read")]
+    jsonrpc: JsonSchemaVersion,
```
benbrandt: Maybe we aren't strict on incoming?

rust/acp.rs
```diff
@@ -169,8 +177,8 @@ fn is_none_or_null<T: Serialize>(opt: &Option<T>) -> bool {
 }
 
 #[derive(Serialize)]
-pub struct JsonRpcMessage<Req, Resp> {
-    pub jsonrpc: &'static str,
+struct OutJsonRpcMessage<Req, Resp> {
+    jsonrpc: JsonSchemaVersion,
```
benbrandt, line 179: Seems we already had this type, but weren't using it. Easy fix

PR #24, opened 2025-08-26
Rewind support

ConradIrwin:
Adds app support for rewind.

This is currently limited to the 1st party agent in Zed, but it would be nice to support it over acp.

### DIFF
--- a/docs/protocol/prompt-turn.mdx
+++ b/docs/protocol/prompt-turn.mdx
@@ -337,15 +337,15 @@ If the agent receives this request while generating, it may also need to cancel.
   "method": "session/rewind",
   "params": {
     "sessionId": "sess_abc123def456",
-    "promptId": "prompt_1234567890",
+    "promptId": "prompt_1234567890"
   }
 }
 ```
 
 <ParamField path="sessionId" type="SessionId">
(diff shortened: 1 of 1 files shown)


kamalmarhubi:
@ConradIrwin will this allow "rewinding" to a prompt id from a different fork of the conversation? If so, should that be a documented requirement of the feature?

Example use cases:
- I rewound by accident and want to return to where I was
- I rewound and tried a different direction, but realised the previous direction was better
- I rewound to free up a bit of space in the context window to answer a question, and then want to continue where I left off with that knowledge

The last comes up for me when I know that ~10 messages back the agent already had enough context to answer a question, but I also know I'm running out of space in the context window to finish my current task.

If, eg, Claude Code supports that kind of "rewind", then this new capability could make it a documented requirement. This would allow ACP client applications to experiment with context branching UIs, even if agents are not directly exposing the forking history.

(This might warrant a rename, as well as some longer-horizon protocol thinking though.)

ConradIrwin:
@kamalmarhubi Potentially, though even the Zed built in agent doesn't support that yet, it sounds like a useful feature.

I don't think anything in the protocol would need to change, but we would need to build it :D

PR #61, opened 2025-09-03
docs: Add yetone/avante.nvim as supported editors

Xuanwo:
@yetone's `avante.nvim` is the first editor that implemetned native support for ACP.


morgankrey:
Done!

PR #63, opened 2025-09-04
feat: the first glance for python-sdk

PsiACE:
This PR try to introduce a Python SDK for ACP,

I will polish it until I successfully use it to connect my agent with Zed.


PsiACE:
@cla-bot check

benbrandt:
@PsiACE thanks for trying to figure this out!
Since this is a bit experimental at the moment, would you be ok doing this in another repo, and then we can link people to it when you are ready? But it seems like you are off to a great start!

PsiACE:
https://github.com/PsiACE/agent-client-protocol-python 

I just released version 0.0.1, which includes a mini Swe agent implementation that can be used with Zed. cc @benbrandt 

image

PR #67, opened 2025-09-05
Introduce experimental session modes

agu-z:
Enables the agent to expose a set of modes that the session can be put in at any given point. The main motivation for this is enabling a permission mode selector for Claude Code, allowing you to toggle between "default" and "plan" mode.

Note: This feature is experimental, and may be removed or changed at any point

### DIFF
--- a/rust/agent.rs
+++ b/rust/agent.rs
@@ -281,6 +281,12 @@ pub struct SessionMode {
 #[serde(rename_all = "camelCase")]
 pub struct SessionModeId(pub Arc<str>);
 
+impl std::fmt::Display for SessionModeId {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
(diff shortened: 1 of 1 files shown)


ConradIrwin:
@agu-z Instead of making this stateful, I think it'd be cleaner to just add a `mode: Option<mode>` to `session/prompt`. (Assuming that you can't meaningfully changes this while the model is generating output).

agu-z:
@ConradIrwin yeah, that's a good idea! 

That said, since making this PR I realized that a core feature of plan mode in CC is that it can ask the user to switch modes in the middle of a turn when it thinks it's ready for implementation. So it's not quite stateless. 

I'm gonna try a few approaches for this mode switch on Monday. I think the simplest would be to request the switch via a tool call, but we'd also need a notification from the agent to let the client know that the mode changed. Another option is to add an explicit method to request a mode switch, in which case the notification wouldn't be needed.

PR #74, opened 2025-09-07
docs: add python-sdk

PsiACE:
As a follow-up to #63 

I have implemented a Python SDK for the agent client protocol. 

Additionally, I have implemented two sample clients in repo: 
- the [mini-swe-agent](https://github.com/SWE-agent/mini-swe-agent/) client
- and the echo agent.

Tested on macOS, with zed v0.202.7

image


benbrandt:
Thanks for this! closing in favor of: https://github.com/zed-industries/agent-client-protocol/pull/82

PR #109, opened 2025-09-23
Rust: Allow Agents to hold onto a Client and vice versa

benbrandt:
- Replace ClientSideConnection/AgentSideConnection structs with type aliases to RpcConnection
- RpcConnection::new now accepts a handler factory FnOnce(Rc<Self>) and returns Rc<Self>

Before, there wasn't a way for an Agent to hold onto a way to
communicate with the Client and vice versa.

Now they can. This matches the behavior in the TS SDK

### DIFF
--- a/rust/acp.rs
+++ b/rust/acp.rs
@@ -77,7 +77,6 @@ pub use tool_call::*;
 pub use version::*;
 
 use anyhow::Result;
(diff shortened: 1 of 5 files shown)


benbrandt:
Going to revisit this as an optional API rather than a breaking change

PR #112, opened 2025-09-24
Add goose to docs about available agents

alexhancock:
Per conversations with @benbrandt this is a doc contribution to show how to use acp with our open source agent, goose which supports it natively


alexhancock:
@cla-bot check

PR #182, opened 2025-10-21
Stabilize Model Selection

benbrandt:
Stabilizes the current model selection API. 

This follows the same pattern as session modes, so nothing too surprising here.

What is new from the previous version is the ability for the agent to notify if it has changed the model itself. 

There aren't currently any agents using it, but some have hinted at it or done something similar in the past, especially in fallback scenarios, so allowing the agent to update the client with this information seems like a fair use case, and makes sure it really is the same lifecycle as the session modes.


benbrandt:
We are discussing an alternative approach, moving to draft for now

PR #207, opened 2025-10-28
Add discriminators to all tagged enums in JSON schema

bennetbo:
This adds a `discriminator` property to all objects that use `oneOf`, where the value of `propertyName` identifies the field by which the different variations are distinguished.

E.g.
```json5
"ToolCallContent": {
  ...
  "discriminator": {
      "propertyName": "type"
  },
  ...
}
```

This helps with some generator tooling for various languages.


benbrandt:
Tested this on the TS side, everything seems fine, so going to merge

PR #210, opened 2025-10-29
RFD Draft: Session Config Options

benbrandt:
Proposing the ability to add Agent provided configuration options to a session.

This makes the current pattern used by session modes more flexible, rather than adding additional hard-coded selectors to the protocol.

### DIFF
--- a/docs/rfds/session-config-options.mdx
+++ b/docs/rfds/session-config-options.mdx
@@ -99,7 +99,7 @@ Something like an `InitializeResponse` that looks like:
 }
 ```
 
-When we introduce this, we could also allow for grouped options, in case there are logical sub-headers and groupings for the options in an individual selector. It
+When we introduce this, we could also allow for grouped options, in case there are logical sub-headers and groupings for the options in an individual selector.
 
 ```json
 {


docs/rfds/session-config-options.mdx
```diff
@@ -0,0 +1,240 @@
...
+
+```json
+{
+  "jsonrpc": "2.0",
+  "id": 1,
+  "result": {
+    "sessionId": "sess_abc123def456",
+    "config_options": [
+      {
+        "id": "mode", // this is the unique `key` for communication about which option is being used
```
anna239, line 61: let's maybe add `type` here and support only `select_single` for now

docs/rfds/session-config-options.mdx
```diff
@@ -0,0 +1,240 @@
...
+
+```json
+{
+  "jsonrpc": "2.0",
+  "id": 2,
+  "method": "session/set_config_option",
+  "params": {
+    "sessionId": "sess_abc123def456",
+    "configId": "mode",
+    "optionId": "code"
```
anna239: we might have different types in the future, should we maybe add `value` here? For the `select-single` it'll be 
{
"optionId": ...
}
in future we'll be able to have lists or, say, just String values

anna239: And the same should also apply too currentValue

docs/rfds/session-config-options.mdx
```diff
@@ -0,0 +1,240 @@
...
+```
+
+We would also likely move session modes to be `@deprecated` in favor of this approach. Until it is removed, we may want Agents to support both fields, and then the Client, if it uses the new config options, should only use the config options supplied and not the `modes` field to avoid duplication.
+
+Open questions at the moment are:
+
+- What types of values do we support initially? Single select? Multiselect? Checkbox?
+  - Push back: Multiselect could theoretically be helpful in some cases, but not utilized by any agents at the moment (could change). A checkbox could be represented by a select as well.
+- Does the Client express capabilities of which types of inputs it supports? And the Agent adjusts the options accordingly?
+- Or, does the Client ignore settings options it doesn't support and the User just doesn't see them?
```
anna239: I like this option, I think it would be hard for an agent to adapt anyway.

docs/rfds/session-config-options.mdx
```diff
@@ -0,0 +1,240 @@
...
+- Should we instead move to a more HTML form inspired approach where we use the terms `label`, and `value`?
+- How do we handle the Agent providing more options than the Client can reasonably display?
```
anna239: I think it's a UX decision,  that I think protocol doesn't have to solve

phil65:
Just some thaught here: MCP's elicitation feature uses a restricted subset of JSON Schema when it comes to asking information from the user via client. Would it make sense to align to that?

docs/rfds/session-config-options.mdx
nerzhulart: I'd prefer other names: `configId` -> `optionId`, `optionId` -> `optionValue`and depending on the option type the handling code will handle the value on its own

benbrandt:
@phil65 it's a good point. The enum type here could work: https://modelcontextprotocol.io/specification/draft/client/elicitation#supported-schema-types
I don't love double lists for values + display names but we could consolidate types here.

It wouldn't be exactly the same as it wouldn't follow the larger request/response lifecycle of an elicitation, but we could at least align on the schemas. I'll take a look before this moves from Draft to Preview for sure

benbrandt:
There is also the interesting distinction here in that we want to enforce a default value, in fact it is required, and we want to also send form state each time, which makes it harder to represent as a schema. 

But we could pull out the form state to a separate struct on the side, so you get a schema + current values

phil65:
Yes agreed, perhaps its not a 100% fit for this mechanism, but (slightly offtopic), I think some equivalent to MCP elicitation would be a nice feature for ACP (could potentially even replace the Permission-Request mechanism, which basically is like elicitation, but limited to 4 buttons?)

benbrandt:
Yep agreed this needs some work, and definitely fits with the use case

docs/rfds/session-config-options.mdx
```diff
@@ -0,0 +1,240 @@
...
+
+Something like an `InitializeResponse` that looks like:
+
+```json
+{
+  "jsonrpc": "2.0",
+  "id": 1,
+  "result": {
+    "sessionId": "sess_abc123def456",
+    "config_options": [
```
agu-z: nit: this should be camelCase

docs/rfds/session-config-options.mdx
agu-z: Good idea! We should make this a discriminated union in the schema, so that fields like `options` are required when `type: "select_single"`. In the future, we may have something like `type: "number"` with associated `min`/`max` fields.

docs/rfds/session-config-options.mdx
```diff
@@ -0,0 +1,240 @@
...
+
+The Client can render the options provided, send updated values to the Agent, and also display any changes the Agent made during the course of it's execution (for example, if it changes modes or models because of fallbacks or a change in strategy, so that the user can always see the current state).
+
+Since we are moving to a world in which there are multiple configuration options, some of which may depend on each other, the Agent MUST provide the complete set of configuration options and their current values whenever a change is made. We would tradeoff some extra data being sent to the Client in order to help minimize the amount of state required to be managed by the Client. The Client would submit a new value, and receive back the full state of all configuration options that it can then replace it's current state with and render. So if changing a model means there are no thinking options, or a new option becomes available, or another value needs to change because the values of an option are different, the Agent will reflect this in its response by providing the entire new state (or an error if it is somehow an invalid selection).
```
agu-z, line 43: I really like how powerful and simple this is!

docs/rfds/session-config-options.mdx
```diff
@@ -0,0 +1,240 @@
...
+            "description": "The most powerful model"
+          }
+        ]
+      }
+    ]
+  }
+}
+```
+
+When we introduce this, we could also allow for grouped options, in case there are logical sub-headers and groupings for the options in an individual selector. It
```
agu-z:
```suggestion
When we introduce this, we could also allow for grouped options, in case there are logical sub-headers and groupings for the options in an individual selector.
```

docs/rfds/session-config-options.mdx
```diff
@@ -0,0 +1,240 @@
...
+            "id": "code",
+            "name": "Code",
+            "description": "Write and modify code with full tool access"
+          }
+        ]
+      },
+      {
+        "id": "models",
+        "name": "Model",
+        "currentId": "ask",
```
agu-z: We should probably call this field `value` or `currentValue` in preparation for non-enum config types.

docs/rfds/session-config-options.mdx
agu-z: One simple solution is to have the order of the options represent relevancy according to the Agent, and then we can mention something like this in the docs:

The Client SHOULD gracefully handle an arbitrary number of options, but it is encouraged to make the first few options in the list easily accessible as these are deemed the most relevant by the Agent.

docs/rfds/session-config-options.mdx
```diff
@@ -0,0 +1,240 @@
...
+          "id": "models",
+          "name": "Model",
+          "currentId": "ask",
+          "options": [..]
+        }
+      ]
+    }
+  }
+}
+```
```
agu-z, line 210: I think we could simplify the Client-side implementation by only including the config options state in this notification, and not in the `session/set_config_option`. The rationale being that the Client is gonna _have_ to handle the notification anyway for changes originating from the Agent, so we could have just have one source of these. 

Let me know if you disagree, though.

docs/rfds/session-config-options.mdx
agu-z: Whoops, missed @anna239's [comment](https://github.com/agentclientprotocol/agent-client-protocol/pull/210#discussion_r2473462170)

docs/rfds/session-config-options.mdx
benbrandt: So you're suggesting, if I set an value, I just get a null back and rely on the agent sending an update if other values have changed as a result?

benbrandt:
Thanks for the feedback everyone! I incorporated it and am going to merge in Draft state. Will get it ready for Preview soon

PR #211, opened 2025-10-29
RFD: session/list API for session discovery and management

ahmedhesham6:
Proposes a new session/list endpoint to enable:
- Discovery of existing sessions
- Session history browsing
- Multi-session management
- Better UX for resuming conversations

The endpoint supports:
- Filtering by creation date, access date, and working directory
- Pagination with limit/offset
- Sorting by various criteria
- Optional capability for agents with session persistence

Includes detailed JSON-RPC request/response examples and addresses common questions about alternatives, interaction with session/load, and pagination strategies.

### DIFF
--- a/docs/rfds/sessions-list-api.mdx
+++ b/docs/rfds/sessions-list-api.mdx
@@ -2,6 +2,9 @@
 title: "Session List API"
 ---
 
+Author(s): [@ahmedhesham6](https://github.com/ahmedhesham6)
+Champion: [@benbrandt](https://github.com/benbrandt)
+
 ## Elevator pitch
 
 Add a `session/list` endpoint to the ACP protocol that allows clients to query and enumerate existing sessions from an agent, enabling session management features like session history, session switching, and session cleanup.
@@ -22,6 +25,7 @@ This creates several problems:
 - **Inconsistent behavior** - Different clients handle session management differently, leading to fragmented experiences
 
 The current workaround is for clients to maintain their own session registry, but this:
+
 - Requires persistent storage on the client side
 - Can get out of sync if sessions are created/destroyed outside the client
 - Doesn't work across different client instances or devices
@@ -137,11 +141,13 @@ The agent responds with a list of sessions and cursor pagination metadata:
 #### Response Fields
 
 **Response object:**
+
 - `sessions` (array) - Array of session information objects
 - `hasMore` (boolean) - Whether more results are available after this page
 - `nextContinuationToken` (string, optional) - Pass this token in the next request's `continuationToken` to fetch the next page
 
 **SessionInfo object:**
+
 - `sessionId` (string, required) - Unique identifier for the session
 - `createdAt` (string, required) - ISO 8601 timestamp when session was created
 - `updatedAt` (string, required) - ISO 8601 timestamp of last activity
@@ -179,6 +185,7 @@ Once this feature exists:
    - Resume interrupted work easily
 
 Agents that implement this feature gain:
+
 - Better visibility into active sessions
 - Opportunity to implement session lifecycle policies
 - Foundation for future features like session sharing or collaboration
@@ -251,6 +258,7 @@ Several alternatives were considered:
    - No standard format
 
 The proposed RPC approach is:
+
 - **Consistent with existing protocol design** - Uses same RPC patterns as other endpoints
 - **Flexible** - Supports filtering, pagination, and agent-specific metadata
 - **Optional** - Agents can opt-in based on their architecture
@@ -259,6 +267,7 @@ The proposed RPC approach is:
 ### Why not make this mandatory for all agents?
 
 Many agents may not have persistent storage or multi-session capabilities. Making this optional:
+
 - Allows simple, stateless agents to remain compliant
 - Reduces implementation burden
 - Lets agents evolve session management over time
@@ -266,6 +275,7 @@ Many agents may not have persistent storage or multi-session capabilities. Makin
 ### How does this interact with `session/load`?
 
 `session/load` remains the mechanism to actually restore a session. `session/list` is for discovery only:
+
 1. Client calls `session/list` to get available sessions
 2. User selects a session
 3. Client calls `session/load` with the chosen `sessionId`
@@ -275,10 +285,10 @@ Agents may support `session/list` without supporting `session/load` (e.g., for r
 ### Should we include session content in the list response?
 
 No, for several reasons:
+
 - **Performance** - Full conversation history could be large
 - **Privacy** - Listing sessions might be less sensitive than exposing full content
 - **Separation of concerns** - Use `session/load` to get full session content
-
 
 ### What about session deletion?
 
@@ -287,6 +297,7 @@ Session deletion is intentionally left out of this RFD to keep scope focused. A 
 ### How should pagination work for large session lists?
 
 We use cursor-based pagination with an opaque continuation token:
+
 - Request includes `limit` and optional `continuationToken`
 - Response includes `hasMore` and `nextContinuationToken` when more data is available
 - Clients should not assume anything about the token's contents; treat it as opaque
@@ -316,7 +327,9 @@ Corresponding response example:
   "jsonrpc": "2.0",
   "id": 2,
   "result": {
-    "sessions": [ /* ... */ ],
+    "sessions": [
+      /* ... */
+    ],
     "hasMore": true,
     "nextContinuationToken": "offset:20"
   }
--- a/docs/docs.json
+++ b/docs/docs.json
@@ -88,7 +88,7 @@
         "tab": "RFDs",
         "pages": [
           "rfds/about",
-          { "group": "Draft", "pages": [] },
+          { "group": "Draft", "pages": ["rfds/sessions-list-api"] },
           { "group": "Preview", "pages": [] },
           { "group": "Accepted", "pages": ["rfds/introduce-rfd-process"] },
           { "group": "Completed", "pages": [] }


docs/rfds/sessions-list-api.mdx
```diff
@@ -0,0 +1,305 @@
...
+- `total` (number) - Total number of sessions matching the filter criteria (for pagination)
+
+**SessionInfo object:**
+- `sessionId` (string, required) - Unique identifier for the session
+- `createdAt` (string, required) - ISO 8601 timestamp when session was created
+- `lastAccessedAt` (string, required) - ISO 8601 timestamp of last activity
+- `cwd` (string, required) - Working directory for the session
+- `title` (string, optional) - Human-readable title (may be auto-generated from first prompt)
+- `preview` (string, optional) - Snippet of the first user message or conversation preview
+- `metadata` (object, optional) - Agent-specific metadata (e.g., message count, error status, tags)
```
yordis: I think it is called `_meta`

docs/rfds/sessions-list-api.mdx
```diff
@@ -0,0 +1,305 @@
...
+#### Request Parameters
+
+All parameters are optional:
+
+- `cwd` (string) - Filter sessions by working directory
+- `createdAfter` (string) - ISO 8601 timestamp, only return sessions created after this time
+- `createdBefore` (string) - ISO 8601 timestamp, only return sessions created before this time
+- `lastAccessedAfter` (string) - ISO 8601 timestamp, only return sessions accessed after this time
+- `limit` (number) - Maximum number of results to return (default: 50, max: 1000)
+- `offset` (number) - Number of results to skip for pagination (default: 0)
```
yordis: my concern is that this is enforcing limit/offset pagination style, sometimes people prefer different strategies based on their storage needs and optimizations of it, such as cursor, or keyset paginations.

ahmedhesham6: What about the usage of continuation token?

Use an opaque continuation token that supports any pagination strategy:

 Request:
```
 {
   "jsonrpc": "2.0",
   "id": 2,
   "method": "session/list",
   "params": {
     "cwd": "/home/user/project",
     "createdAfter": "2025-10-20T00:00:00Z",
     "limit": 20,
     "continuationToken": "eyJsYXN0SWQiOiJzZXNzXzEyMyIsInRzIjoxNjk4NzY1NDMyfQ==",
     "sortBy": "lastAccessedAt",
     "sortOrder": "desc"
   }
 }
```

 Response:
```
 {
   "jsonrpc": "2.0",
   "id": 2,
   "result": {
     "sessions": [ /* ... */ ],
     "hasMore": true,
     "nextContinuationToken": "eyJsYXN0SWQiOiJzZXNzXzQ1NiIsInRzIjoxNjk4NzYwMDAwfQ=="
   }
 }
```

And the Agent is responsible for the token and can choose his strategy

```
 //Offset-based:
 continuationToken = base64({"offset": 20})

 //Cursor-based:
 continuationToken = base64({"lastEvaluatedKey": {...}})

 //Keyset:
 continuationToken = base64({"lastId": "sess_123", "lastAccessedAt": "2025-10-29T14:22:15Z"})
```

benbrandt: I would even say at the protocol level we could leave this even more open-ended, we could allow for either a number or a string. It is really just a value that the agent can provide the client so they can send it back, and then it is up to the agent to encode/decode that information as they wish.

docs/rfds/sessions-list-api.mdx
```diff
@@ -0,0 +1,306 @@
...
+- `sortBy` (string) - Field to sort by: `"createdAt"`, `"lastAccessedAt"`, or `"title"`
+- `sortOrder` (string) - Sort direction: `"asc"` or `"desc"` (default: `"desc"`)
```
benbrandt: I wonder how much value we gain from the sorting with the tradeoff of extra complexity we add to every Agent implementation.

I wonder if in most cases, if I can have a list of sessions from newest to oldest, and I have search, is that enough as a starting point?

docs/rfds/sessions-list-api.mdx
```diff
@@ -0,0 +1,306 @@
...
+    "total": 47
+  }
+}
+```
+
+#### Response Fields
+
+**Response object:**
+- `sessions` (array) - Array of session information objects
+- `total` (number) - Total number of sessions matching the filter criteria (for pagination)
```
benbrandt: Is this number different than the number of elements in the sessions array?

docs/rfds/sessions-list-api.mdx
```diff
@@ -0,0 +1,306 @@
...
+
+**SessionInfo object:**
+- `sessionId` (string, required) - Unique identifier for the session
+- `createdAt` (string, required) - ISO 8601 timestamp when session was created
+- `lastAccessedAt` (string, required) - ISO 8601 timestamp of last activity
```
benbrandt: Do we want last accessed? Or updatedAt?

docs/rfds/sessions-list-api.mdx
```diff
@@ -0,0 +1,306 @@
...
+1. **Discover existing sessions** - Clients cannot query what sessions exist on an agent
+2. **Display session history** - Users cannot see a list of their past conversations
+3. **Manage multiple sessions** - Switching between sessions requires clients to track session IDs themselves
+4. **Clean up old sessions** - No way to discover stale or abandoned sessions for cleanup
+
+This creates several problems:
+
+- **Poor user experience** - Users cannot browse their conversation history or resume previous sessions easily
+- **Client-side complexity** - Each client must implement its own session tracking and persistence
+- **Session orphaning** - Sessions may be left running or consuming resources without the client knowing
```
benbrandt: I don't see how this is related? 
Unless this list somehow also returns which sessions the agent currently has in memory/running?

docs/rfds/sessions-list-api.mdx
```diff
@@ -0,0 +1,306 @@
...
+   - `listSessions` capability flag
+
+2. **Update protocol documentation** in `/docs/protocol/session-setup.mdx`:
+   - Document the new endpoint
+   - Explain when to use it vs. maintaining client-side session tracking
+   - Provide examples of common use cases
+
+### Phase 2: Reference Implementation
+
+3. **Implement in Rust SDK** (`rust/agent.rs` and `rust/client.rs`):
```
benbrandt, line 210: Note, these are just types that actually generate the schema. There is no trait in here anymore, it lives in the rust-sdk repo now

docs/rfds/sessions-list-api.mdx
```diff
@@ -0,0 +1,306 @@
...
+### Compatibility Considerations
+
+- **Backward compatible**: Existing agents continue working without implementing this
+- **Capability-based**: Clients check for `listSessions` capability before using
+- **No breaking changes**: No modifications to existing endpoints
+
+### Security Considerations
+
+- **Session isolation**: Agents must ensure sessions are only listed for the authenticated client
+- **Privacy**: The `preview` field should be optional and respect any privacy settings
```
benbrandt: To this point, we could also start without the `preview` field and add it later?

docs/rfds/sessions-list-api.mdx
```diff
@@ -0,0 +1,306 @@
...
+
+The proposed RPC approach is:
+- **Consistent with existing protocol design** - Uses same RPC patterns as other endpoints
+- **Flexible** - Supports filtering, pagination, and agent-specific metadata
+- **Optional** - Agents can opt-in based on their architecture
+- **Simple** - Single request/response pattern, easy to implement and use
+
+### Why not make this mandatory for all agents?
+
+Many agents may not have persistent storage or multi-session capabilities. Making this optional:
```
benbrandt, line 269: This would also be a breaking change 😄

benbrandt (commented):
Hi thanks for putting in a lot of work on this! I think it's a great start. 
I left some notes here, curious to get your feedback. Either way, I think we could bring this to `Draft` state soon and do some more work before bring it to `Preview` for final review

docs/rfds/sessions-list-api.mdx
ahmedhesham6: I agree

docs/rfds/sessions-list-api.mdx
ahmedhesham6: Yeah maybe it extra complex as a starting point, it just offloads the flexibility to the client so each client can make the best experience, but if there isn't any clear upside from having it, then lets start small

docs/rfds/sessions-list-api.mdx
ahmedhesham6: Yeah its the total number of sessions for the pagination, but we already agreed to update the pagination approach so it will be replaced with `hasMore`

docs/rfds/sessions-list-api.mdx
ahmedhesham6: Yeah this should be removed

docs/rfds/sessions-list-api.mdx
ahmedhesham6: Yeah that's why its optional, do we want it to be required?

benbrandt: Ha no, just a reason not to make it mandatory. as you pointed out there are valid reasons for an agent not to support it

docs/rfds/sessions-list-api.mdx
benbrandt: OK @bennetbo  pointed out that MCP already has a nice pattern around this that I think should work for our session list as well:

https://modelcontextprotocol.io/specification/draft/server/utilities/pagination

There are pretty minimal requirements on the Agent side, and I think we can drive the initial UX we need on the Client side with something like this.

yordis: opaque field is the way to go for sure, let the server to decide what it is inside

docs/rfds/sessions-list-api.mdx
```diff
@@ -0,0 +1,334 @@
...
+  }
+}
+```
+
+#### Response Fields
+
+**Response object:**
+- `sessions` (array) - Array of session information objects
+- `hasMore` (boolean) - Whether more results are available after this page
+- `nextContinuationToken` (string, optional) - Pass this token in the next request's `continuationToken` to fetch the next page
```
benbrandt, line 147: What do you think of following MCP's pattern here?: https://modelcontextprotocol.io/specification/2025-06-18/server/utilities/pagination

They call it `nextCursor` and you use the presence of it for the same meaning as your hasMore boolean, which I think makes sense

benbrandt:
I think this is already well on it's way to "Preview" stage. In the spirit of the process, I am going to go ahead and merge this as-is as a "draft" and we can keep iterating on it for getting it ready for "Preview" and then we can "Accept" and start working on it.

Thanks so much for the work here!

benbrandt:
Draft published here: https://agentclientprotocol.com/rfds/session-list

PR #215, opened 2025-10-31
Add cagent to the list of ACP agents

rumpl:


rumpl:
@benbrandt I would be happy to chat! Shoot me an email you find on my github account ;)

PR #219, opened 2025-10-31
docs: Move brand assets to the repo

danilo-leal:
Have them be stored here instead of in the Zed site repo.


docs/assets/acp-brand.zip
maxdeviant, line 1: Oof, that’s a big binary asset 😭

danilo-leal: 475kb? 🤔

maxdeviant: Yea, but any time we make a change to the contents of the assets in the zip, it will add another 475kb to the size of the Git repository.

We could easily set up a CDN for ACP and serve binary assets from there.

PR #225, opened 2025-11-05
RFD: Limit filter capabilities for session/list

benbrandt:
Let's start with something a bit simpler for a first pass. I'd like to
verify the basic implementation a bit in a few agent before committing
to all of these filters.

Limit seems like a fair (and easy to implement one) so let's leave it.


yordis (commented):
I agree with the strategy to keep it limited and add as we learn, the spec an always opt-in to use `_meta` to test the water, and then report back the learnings, so it shouldn't be a "blocker"

PR #226, opened 2025-11-06
Feat: Add session/list API for managing and discovering sessions

ahmedhesham6:
This commit introduces the `session/list` capability, allowing agents to list existing sessions with optional filtering and pagination. It includes the `ListSessionsRequest` and `ListSessionsResponse` structures, along with the necessary updates to the protocol documentation and JSON schema. This feature is marked as **UNSTABLE** and may change in future iterations.

### DIFF
--- a/docs/protocol/schema.unstable.mdx
+++ b/docs/protocol/schema.unstable.mdx
@@ -202,6 +202,9 @@ Only available if the Agent supports the `listSessions` capability.
 
 **Properties:**
 
+<ResponseField name="_meta" type={"object"}>
+  Extension point for implementations
+</ResponseField>
 <ResponseField name="cursor" type={"string | null"}>
   Opaque cursor token from a previous response's nextCursor field for
   cursor-based pagination
@@ -2464,9 +2467,6 @@ Information about a session returned by session/list
 
 <ResponseField name="_meta" type={"object"}>
   Extension point for implementations
-</ResponseField>
-<ResponseField name="createdAt" type={"string | null"}>
-  ISO 8601 timestamp when session was created
 </ResponseField>
 <ResponseField name="cwd" type={"string"} required>
   The working directory for this session. Must be an absolute path.
--- a/docs/rfds/session-list.mdx
+++ b/docs/rfds/session-list.mdx
@@ -39,7 +39,6 @@ Add a new `session/list` JSON-RPC method to the protocol that returns metadata a
    - `sessionId` - Unique identifier
    - `cwd` - Working directory for the session
    - `title` - Optional human-readable title (could be auto-generated from first prompt)
-   - `createdAt` - Timestamp when session was created
    - `updatedAt` - Timestamp of last update to the session
    - `_meta` - Optional agent-specific metadata
 2. **Support filtering and pagination**:
@@ -98,7 +97,6 @@ The agent responds with a list of sessions and cursor pagination metadata:
     "sessions": [
       {
         "sessionId": "sess_abc123def456",
-        "createdAt": "2025-10-29T10:30:00Z",
         "updatedAt": "2025-10-29T14:22:15Z",
         "cwd": "/home/user/project",
         "title": "Implement session list API",
@@ -109,14 +107,12 @@ The agent responds with a list of sessions and cursor pagination metadata:
       },
       {
         "sessionId": "sess_xyz789ghi012",
-        "createdAt": "2025-10-28T09:15:00Z",
         "updatedAt": "2025-10-28T16:45:30Z",
         "cwd": "/home/user/another-project",
         "title": "Debug authentication flow"
       },
       {
         "sessionId": "sess_uvw345rst678",
-        "createdAt": "2025-10-27T14:00:00Z",
         "updatedAt": "2025-10-27T15:30:00Z",
         "cwd": "/home/user/project"
       }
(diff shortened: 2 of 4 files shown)


schema/schema.json
```diff
@@ -1759,6 +1816,36 @@
...
+        "cwd": {
+          "description": "Working directory for the session",
+          "type": "string"
+        },
+        "sessionId": {
+          "$ref": "#/$defs/SessionId",
+          "description": "Unique identifier for the session"
+        },
+        "title": {
+          "description": "Human-readable title (may be auto-generated from first prompt)",
```
codefromthecrypt: Made a discussion comment [here](https://github.com/orgs/agentclientprotocol/discussions/60#discussioncomment-14936950), but wondering

While true it may be auto-generated, the impact is you shouldn't cache this right? Because it might be initially empty or a fake initial value like "Goose ACP". It seems several flows indeed do a lazy one-time update, but I'm not sure if that's only one time always (e.g. should this just be assumed never cacheable unlike the sessionId, which should me immutable?)

schema/schema.json
```diff
@@ -1759,6 +1816,36 @@
...
+          "type": ["string", "null"]
+        },
+        "updatedAt": {
+          "description": "ISO 8601 timestamp of last activity",
```
codefromthecrypt: This might be a great place to clarify the impact of the other field.

```suggestion
          "description": "ISO 8601 timestamp of last activity, such as title update",
```

schema/schema.json
```diff
@@ -1759,6 +1816,36 @@
...
+          "description": "Working directory for the session",
+          "type": "string"
+        },
+        "sessionId": {
+          "$ref": "#/$defs/SessionId",
+          "description": "Unique identifier for the session"
+        },
+        "title": {
+          "description": "Human-readable title (may be auto-generated from first prompt)",
+          "type": ["string", "null"]
```
codefromthecrypt, line 1921: I'm not sure if we have value both marking something optional (not required field) and also nullable. since this is not a required field, should we just simplify to when present, put in a valid string?

```suggestion
          "type": "string"
```

schema/schema.json
ahmedhesham6: There is nothing in the spec says it should be cached, Its more of an ACP client preference but when using `session/list` I agree that we should return the latest title

schema/schema.json
ahmedhesham6: That’s debatable from an experience standpoint. Why should the order of the sessions change just because the title changed? It also depends on each agent’s behavior — some update the session’s `updated_at` when the title changes, and some don’t.

codefromthecrypt: interesting response, and I can see the sense of it. Since many "update" fields are updated unconditionally, would you be ok spelling this out a bit more?

"ISO 8601 timestamp (UTC) set by the agent on meaningful session activity. Trivial or cosmetic changes MAY leave it unchanged."

ignatov:
Hey! Sorry for jumping in here. I was about to propose a feature like `send_session_name_update_event` or something similar. My intention is that the agent can update the session name UI multiple times during its work, and the UI should be updated accordingly.

I think such an addition naturally fits into that PR.
@ahmedhesham6, @benbrandt what do you think?

ahmedhesham6:
Hey @ignatov — I like the idea. I just don’t think it belongs in this PR.

Something like `send_session_name_update_event` feels like a client↔agent notification concern. We should either:

* extend the `session/update` notification schema to carry session metadata (including name), or
* introduce a new notification type specifically for session-name updates.

Either way, that’s a broader protocol/API change and should go through a separate RFD.

benbrandt:
Hi yes I agree:

1. I think we do need this session update 100%
2. I don't think it belongs in this RFD/PR

Mainly because I want these changes to be reviewable, and while this notification is connected, I think it can build on top of it.

One thing we may want to decide if people have strong opinions is if we should call this field "title" or "name" if we have strong opinions there, so that we can have one concept. 

But I know there are some teams that definitely have opinions on titles and how they should be updated (should the client be able to rename it for example) so I think there will be enough to discuss in its own RFD for sure.

schema/schema.json
benbrandt: Hi @codefromthecrypt I understand this may not feel "clean" to you, but I have played with the generators for some of the SDKs, and not having this means that many validators would not accept `null` if it got sent in.

We have seen some languages not always strip these out, and so it feels safer to leave it in.

We could maybe remove it in a later major version of the protocol, but for now I would rather keep this consistent, as removing this could cause some SDKs to start rejecting valid requests.

schema/schema.json
benbrandt: I left a comment on the discussion. I think all of these concerns (for me at least) don't affect the content of this PR

schema/schema.json
benbrandt: Why do we need to be so prescriptive over when the agent does/does not update this?

ignatov:
Okay, in that case SessionInfo should be aligned with the current session primitive. I think it should use the same structure that describes the current session and we use in `sessions/list` call.

benbrandt:
@ignatov I'm confused now... this is the PR that introduces `Sessioninfo` and is the first time we've had such a primitive. Which other one are you referring to?

benbrandt:
@ignatov I am going to go ahead and merge as-is. This RFD is still in draft stage, not preview yet, so we can still continue the discussion if you feel there is something missing here

ignatov:
> I am going to go ahead and merge as-is. This RFD is still in draft stage, not preview yet, so we can still continue the discussion if you feel there is something missing here

@benbrandt, sure

codefromthecrypt:
please tag the session name update PR or issue back here, as it is really important agreed!

PR #233, opened 2025-11-10
RFD: Forking of existing sessions

josevalim:


SteffenDE:
For context: we opened https://github.com/zed-industries/claude-code-acp/pull/145 to explore how this could be implemented. The PR also suggested a `session/resume` functionality, but that would be a separate draft.

For Claude Code, the SDK already provides the blocks we'd need to implement this: `forkSession` and `resume`. I'm imagining the API could look like this:

```json
{
  "jsonrpc": "2.0",
  "method": "session/fork",
  "params": {
    "sessionId": "EXISTING-ACP-SESS-ID"
  }
}
```

Optionally, we could require to also pass any other parameters from `session/new`, like `cwd` and `mcpServers` for the forked session, but that might be different depending on the agent.

The agent would then reply with the same format as `session/new`, creating a new ACP `sessionId` for the forked session.

docs/rfds/session-fork.mdx
```diff
@@ -0,0 +1,66 @@
...
+
+I can also see this feature being extended in the future to support an optional
+message ID, so the fork happens at a specific message, allowing clients to implement
+functionality like editing previous messages and similar.
+
+## Implementation details and plan
+
+> Tell me more about your implementation. What is your detailed implementation plan?
+
+<!--
```
benbrandt: Let's add some basic json inputs here (I think @SteffenDE already had something)

I think starting with just the session id you want to fork makes sense. But yes we could also allow for optionally passing in `cwd` and `mcpServers`? But we'd have to be clear what passing those does (I assume you pass in either it overrides, otherwise it default to current session settings)

And then I assume it would also return with the same output as a `NewSessionResponse`

Another idea might be to have this be an optional parameter on `session/new`. The downside is it is easier to reject an unknown method, whereas an unknown parameter would just silently fail because the `new` method would ignore it.

Maybe something to consider as a protocol v2 consolidation, but I think for now, having a new method makes sense.

Either way we should also add a capability. If we feel there is more we want to expand on here, we could do `forkSession: {}` or we could do the simpler `forkSession: bool`

I really wish we had created a `session: { load: bool, fork: bool }` setup for capabilities... I am also open to doing that and figuring out a deprecation path... but maybe not worth it right now.

docs/rfds/session-fork.mdx
```diff
@@ -0,0 +1,66 @@
...
+    Use this section to add details that were not covered in the "What we propose to do about it" section and also include an implementation plan with phases.
+
+    Note: This section is OPTIONAL and NOT RECOMMENDED when RFDs are first opened. It can distract from the discussion of the problem.
+-->
+
+## Frequently asked questions
+
+> What questions have arisen over the course of authoring this document or during subsequent discussions?
+
+<!--
```
benbrandt: Maybe we add something here, since I asked it, about why we are creating a new method?

benbrandt (commented):
Love it. Added some comments, I can also take a look to figure out what we'd need to do for Codex and others as well

docs/rfds/session-fork.mdx
josevalim: > But yes we could also allow for optionally passing in cwd and mcpServers?

I would probably not allow cwd and mcpServers to be given for the same reasons given here: https://github.com/agentclientprotocol/agent-client-protocol/pull/234#discussion_r2533663521 - which also helps explain why this should be a different method (as it accepts different options than session/new).

anna239: I believe quite often you'd like to fork not from the last message, but from some point in-between. So for forking we need not only sessionId but also some checkPointId

josevalim: Agreed, I will add that note in, we will need to describe via capabilities if forking from a given message/checkpoint is possible or not.

anna239: Or we can just add a checkpoint property to some of the messages (from the top of my ming — prompt response messages, maybe tool confirmations) and if an agent returns checkpoints, it means that it supports forking a session? wdyt?

SteffenDE: > and if an agent returns checkpoints, it means that it supports forking a session?

It would mean that **checkpoints** are supported, right? You could have an agent only supporting forking from the last message, so if it did return checkpoints for each prompt, users could think that you can fork from old ones as well

anna239: Ok, makes sense. Do you see any other application for checkpoints at the moment?

SteffenDE: No, but I agree that it is useful, so including an opt-in way for forking from a checkpoint makes sense for agents where that's easy to implement. As we have the capabilities as part of the initialization, announcing checkpoint support there makes the most sense to me, even if the presence of a `checkpointId` itself would be sufficient.

benbrandt: I think we should also figure out message ids as well #244 
But it makes sense that the agent might want control over a where the forking could happen.

However, if so, wouldn't we need to also check for this on the last message? Like, that last message would need to support forking with a checkpointId, no?

If not, then I wonder if just a messageId would suffice

josevalim: @benbrandt Claude Code does not currently support specifying a message, so I think we need to have a general "fork wherever you are right now" and we reserve the messageId as an additional capability. So my goal is that this RFD does not specify the messageId/checkpointId yet but we design it well enough for future extensions by allowing the forkSessions capability to be an object with future keys. Thoughts?

benbrandt: Yeah I think that makes sense. I already regret using bools anywhere in the capabilities haha.

I think if an agent has more restrictions than "most recent message" then they wouldn't be able to support forking over ACP (yet) and we could add other ways of forking in the future

josevalim: @benbrandt if we want to be super safe, we could pass `{ latest: true }` as argument to session/fork (and also have it defined in the capability). For example, if we had done so for session/load, then session/resume could have been a more natural extension to session/load by passing another option.

Although, on the other hand, we could argue one method with three distinct options vs three methods with no options are all the same, especially if the return types are different (as in session/resume and session/load returning different things).

We could also call it `session/forkLatest` but, all in all, `session/fork` is likely fine.

benbrandt: I think it's fine. To be honest, I feel most agents should be able to fork from an end turn state. Like that should be a fairly valid state, and if not, it's probably because you got an error on that turn.

So I'd be fine with it defaulting to the latest, and then we can add an optional checkpoint anchor later on if you want to fork from a different spot

josevalim: So, in my humble opinion, this is good to go! What is the next step, assuming you folks also agree? :)

benbrandt (approved):
I made some small tweaks to inline some examples, and also tweaked the capability structure a bit since we have a few session capabilities now (and could also make a plan to migrate the load one)

But looks good to me!

josevalim:
Thank you!!! 🥳 🍾

faern:
> And the agent would respond with optional data such as config options, the same as `session/load`.

I think the `session/fork` specification should be stricter on how it should mimic `session/load`. Currently the `session/fork` proposal does not specify that an agent MUST stream all previous `session/update` notifications on fork, but `session/load` require that.

A real world problem with this is that both OpenCode and goose implement experimental forking, but only the former stream all pre-fork notifications to the client, goose does not replay any previous messages at all. It would be great if the specification could make this consistent and unambiguous.

PR #234, opened 2025-11-10
RFD: Resume existing sessions

josevalim:
This document proposes the addition of a 'session/resume' command to allow users to resume existing sessions without returning previous messages, addressing the limitations of the current 'session/load' command.

### DIFF
--- a/docs/updates.mdx
+++ b/docs/updates.mdx
@@ -4,6 +4,13 @@ description: Updates and announcements about the Agent Client Protocol
 rss: true
 ---
 
+<Update label="November 26, 2025" tags={["RFD"]}>
+## session/resume RFD moves to Draft stage
+
+The RFD for adding a "session/resume" method to the protocol has been moved to Draft stage. Please review the [RFD](./rfds/session-resume) for more information on the current proposal and provide feedback as work on the implementation begins.
+
+</Update>
+
 <Update label="November 20, 2025" tags={["RFD"]}>
 ## $/cancelRequest RFD moves to Draft stage
 


docs/rfds/resume-session.mdx
```diff
@@ -0,0 +1,100 @@
...
+> What are you proposing to change?
+
+We propose adding the ability to resume existing sessions. This is similar to session/load,
+except it does not return previous messages.
+
+## Status quo
+
+> How do things work today and what problems does this cause? Why would we change things?
+
+While the spec provides a "session/load" command, most coding agents do not implement it.
```
gsabran: I'm wondering what was the intent to have `session/load` if main coding agents don't support it.
If it is indeed the case that `session/load` is not the correct abstraction for the time being, should ACP drop it entirely in favor of the `session/resume` that is more aligned with current coding agent capabilities?

benbrandt: I'm open to both actually. As laid out `resume` is lighter weight, but it comes with tradeoffs that the client has to either store the entire previous conversation, or the user is left without the history of what happened before.

The goal of `load` was that the Agent could be in charge of persistence, but obviously this takes some more work on the Agent side.

We haven't gotten to pulling our zed agent out into it's own process, but when we do this would be how we would handle loading the session, we'd do it from the agent not the client side, which is where the original design came from.

docs/rfds/resume-session.mdx
```diff
@@ -0,0 +1,100 @@
...
+This means that, once you close your editor, browser, etc, you can't resume the conversation.
+
+An obvious answer to this question would be to implement "session/load" for existing coding
+agents. For example, `claude-code-acp` could intercept all messages, write them to disk,
+and emulate "session/load". We considered contributing this when we realized that, we already
+store all messages in our own application, independent from ACP, given we want to provide
+users to navigate or share old chats even if the coding agent is removed. And if we were to use
```
benbrandt: This is an interesting point. I know there is some work by other teams as well of putting a layer between the agent and the client (a proxy or something similar) that could store these events across agents. You'd still need to deal with persistence, but this could potentially be abstracted across agents and also allow for replay across multiple clients (i.e. I want to kick a session off on my computer and then load/interact with it on my phone).

I think the use case you have is definitely valid, I just wonder who is best suited to own the state. I'm biased towards the agent (or something sitting in front of the agent) but I am open to changing my mind on this.

(cc @nikomatsakis if you have thoughts)

docs/rfds/resume-session.mdx
```diff
@@ -0,0 +1,100 @@
...
+"session/load", we would likely just discard them anyway.
+
+In those cases, the ability to resume a session is beneficial because:
+
+* it is exactly what we need
+* it is more straight-forward to implement for both clients and servers
```
benbrandt: It is more straightforward, assuming the client doesn't have the responsibility to also implement storing all conversations.

One reason this causes some issues is if you have a session on the client that got deleted on the agent, are you actually able to resume it? Would we instead have some way to "seed" a conversation from the client, passing the ACP messages through and the agent picks it up from there?

I think if we move more data ownership to the client, we may need to think about allowing for that sort of data flow.

But if the client accepts resuming without a previous history (maybe only showing the title and/or summary available from a list method) it is certainly easy to implement.

docs/rfds/resume-session.mdx
```diff
@@ -0,0 +1,100 @@
...
+
+## What we propose to do about it
+
+> What are you proposing to improve the situation?
+
+Two possible options:
+
+* Add a "session/resume" command
+* Add an option to "session/load" which allows us to continue even if
```
benbrandt: This second option would basically change the language from `load` from:

"The Agent **MUST** replay the entire conversation to the Client"

to 

"The Agent **SHOULD** replay the entire conversation to the Client"

Which obviously from the implementation side isn't _that_ different in that a list of 0 events is processed the same as a list of X events. 

But obviously the UX concerns are quite different.

docs/rfds/resume-session.mdx
```diff
@@ -0,0 +1,100 @@
...
+    Keep this section up-to-date as discussion proceeds. The goal is to capture major points that came up on a PR or in a discussion forum -- and if they reoccur, to point people to the FAQ so that we can start the dialog from a more informed place.
+-->
+
+### What alternative approaches did you consider, and why did you settle on this one?
+
+The biggest question is if it makes sense to support both "session/load" and
+"session/resume". However, I do wonder: how portable "session/load" is in practice?
+
+When we start a new session over ACP, we introduce custom MCP tools and configuration.
+This means that, while we could use "session/load" to load our own chats and chats
```
benbrandt: Yeah I guess the issue here is that we allow you to load a session with a different cwd and mcp servers, which indeed was maybe a mistake here reflecting on it.

docs/rfds/resume-session.mdx
```diff
@@ -0,0 +1,100 @@
...
+not started by us, I presume loading third-party chats would lead to a flawed user
+experience, as our tools would not be available. And depending on the ACP implementation,
+not even the capabilities would be respected.
```
benbrandt: I'm not sure why capabilities wouldn't be respected? Can you elaborate here?

docs/rfds/resume-session.mdx
```diff
@@ -0,0 +1,100 @@
...
+
+Therefore, I see functionality such as "session/load" and "session/list" being
+useful for those who want to provide a different interface for coding agents, without
+injecting custom functionality. Outside of that, the other benefit of "session/load"
+is that we could rely on the coding agent's own storage, but unfortunately that does
+not seem to be exposed at the moment. So if we have to emulate "session/load" at the
```
benbrandt: I would say from the agents where we apply adapters this is try, it isn't exposed.
However for the teams who are building out native ACP integration, I wonder how true this statement holds up. 
The agent would have access to its own storage and I would assume be able to trivially remap the storage back to ACP events.

docs/rfds/resume-session.mdx
```diff
@@ -0,0 +1,100 @@
...
+ACP layer, we may know have chats stored in the coding agent, in ACP, and in our
+applications.
+
+If we can trivially implement "session/load" across agents, then I don't believe this
+RFD has weight. But assuming we don't, we may need to deal with these trade-offs:
+
+* If "session/resume" exists, coding agents may avoid implementing "session/load",
```
benbrandt: It seems like a reasonable fallback. If an agent supports the others, it makes for a nicer experience, and then it is a matter of how much the client wants to handle fallbacks for resume I guess.

docs/rfds/resume-session.mdx
josevalim: >  (i.e. I want to kick a session off on my computer and then load/interact with it on my phone)

In this case it feels we need a completely different "session/load" implementation, which is one that accepts an existing chat as input (or perhaps "session/replay"), but even that comes with some caveats (which I will mention below).

docs/rfds/resume-session.mdx
josevalim: Right, in our experience changing the tools or cwd of existing chats doesn't work in practice, because the whole context will be filled with paths and tools that may no longer be available, and the agent will try to use them anyway.

docs/rfds/resume-session.mdx
josevalim: If capabilities are implemented as custom MCP tools, then it most likely won't work, because the agent will see the old tools in the context/history and try to use them. So for example, a Claude Code chat using "Read" tools, which are then replaced by `mcp_acp_read`, won't work as expected.

docs/rfds/resume-session.mdx
josevalim: > One reason this causes some issues is if you have a session on the client that got deleted on the agent, are you actually able to resume it? Would we instead have some way to "seed" a conversation from the client, passing the ACP messages through and the agent picks it up from there?

If the chat is deleted on the agent, then you can't session/resume it, but you also can't session/load it anyway.

We could provide some sort of seeding/replay (if we did, then we wouldn't need to support something like fork!) but it is worth saying replaying conversations across different agents doesn't really work as the tools would have different names anyway. Even if you are using the same agent, across machines, then you need to guarantee they have very similar configurations. Even loading chats across different models could be problematic, especially if they have different instructions in relation to how diffs must be built and applied (the system prompt will say one thing, history will say something else).

So overall, I think seeding/replaying chats is full of pitfalls. Once you get the data out, it is pretty much read-only.

josevalim:
@benbrandt thank you for the feedback. Our discussion made me think about another angle we could approach "session/resume": if we have "session/resume", then we can implement "session/load" through a proxy that stores the data and converts "session/load" into "session/resume" operations. With this in mind, I am convinced "session/resume" is the primitive we want to aim for, and "session/load" is an optional affordance in case the agent exposes its own data.

I was on the fence about this proposal but, after this batch of feedback, I feel confident it is the right way forward!

docs/rfds/resume-session.mdx
benbrandt: This is a different case that comes up more with hosted/remote agents.

Then it can actually be the same agent session you are connecting to, but you need a way to load the previous conversation

docs/rfds/resume-session.mdx
benbrandt: Yep fair, my point is I guess the more we assume the Client is storing a bunch of information there needs to be some way (ideally) for the Agent to indicate that the session is no longer available so that resuming it isn't available.

This could be handled by an agent supporting both `list` and `resume` so the Client at least knows if it is resumable or not

docs/rfds/resume-session.mdx
benbrandt: Yeah ... I'll have a think about how we can message this for load in v1 and probably remove in v2

docs/rfds/resume-session.mdx
benbrandt: Ok sorry I was thinking about ACP capabilities in which case I didn't see the issue. But now I see your main concern is in invalid context in the session, which makes sense

benbrandt:
@josevalim agreed. I think obviously there are richer alternatives, but having at least a "resume" as a minimal affordance would already be a win for Zed in terms of the current failure case of having no way to pick up a session again.

We could at least show the fact that we can't show previous history, but you could at least pick up knowing where you left off.

I think it's a nice primitive and I think composed with something like #243 there could be some nice shared primitives for agents who don't implement this themselves

docs/rfds/resume-session.mdx
anna239: We were thinking about the remote scenario, and for that case it looks like we need different session/load strategies: loading last N messages, loading only `important` ones (like input requests in so on) Could loading a session without any messages be considered just one of such strategies?

josevalim: @anna239 could you share more information about the remote scenario and its challenges, so I can best understand how it relates to session/load and session/resume? Thank you :heart:

anna239: In remote scenario connection can be lost or you may want to connect to the existing session with a new client. In this cases you don't always want to load all messages, you might want to load starting from some timestamp or, f.i. load only messages that require user's attention — errors, input requests and so on.

josevalim: Got it! I think it is fine if resume is a flag to session/load. I think we can argue both ways! The only question is how to express these via capabilities, given loadSession is a Boolean? But that’s a problem that needs to be solved for your case anyway.

benbrandt:
Thinking about this some more, I think having this as a separate method provides a few benefits:

- for clients that for whatever reason don't want the history, they can just resume
- for agents that can only supply resume, a proxy on top could provide load on top of it, but the agent is still clear on what it supports
- for agents who support both, it should be trivial to use the same resume functionality, they just either replay events or do not

By having it as separate methods it is also clear which behavior you'll get.

@anna239 do you have any thoughts on this?

anna239:
@benbrandt agree, let's go this way and we'll be able to add additional parameters to session/load later if we decide to do so, but current solution will provide some instant benefits

josevalim:
@benbrandt I have updated the FAQ and the proposal with the latest feedback. In my opinion, this is good to go. The only open question is if the capability is `resumeSession: boolean` or `resumeSession: {}`.

PR #244, opened 2025-11-17
RFD: introduce `messageId` field

Nemtecl:
This document proposes the addition of a `messageId` field so that we can identify user and assistant message

### DIFF
--- a/docs/rfds/message-id.mdx
+++ b/docs/rfds/message-id.mdx
@@ -6,7 +6,7 @@ Author(s): [@michelTho](https://github.com/michelTho), [@nemtecl](https://github
 
 ## Elevator pitch
 
-Add a `messageId` field to `agent_message_chunk`, `user_message_chunk`, and `agent_thought_chunk` session updates, to `session/prompt` requests and responses, to uniquely identify individual messages within a conversation. Both clients and agents can generate message IDs using UUID format. This enables clients to distinguish between different messages beyond changes in update type and lays the groundwork for future capabilities like message editing and session deduplication.
+Add a `messageId` field to `agent_message_chunk`, `user_message_chunk`, `agent_thought_chunk` session updates and `session/prompt` requests, and a `userMessageId` field to `session/prompt` responses, to uniquely identify individual messages within a conversation. Both clients and agents can generate message IDs using UUID format. This enables clients to distinguish between different messages beyond changes in update type and lays the groundwork for future capabilities like message editing and session deduplication.
 
 ## Status quo
 
@@ -59,7 +59,7 @@ As an example, consider this sequence where a Client cannot reliably determine m
 
 ## What we propose to do about it
 
-Add a `messageId` field to `AgentMessageChunk`, `UserMessageChunk`, and `AgentThoughtChunk` session updates, and to the `session/prompt` response. This field would:
+Add a `messageId` field to `AgentMessageChunk`, `UserMessageChunk`, `AgentThoughtChunk` session updates and `PromptRequest`, and a `userMessageId` field to `PromptResponse`. These fields would:
 
 1. **Provide stable message identification** - Each message gets a unique identifier that remains constant across all chunks of that message.
 
@@ -89,14 +89,14 @@ When the Client sends a user message via `session/prompt`, it can optionally inc
 }
 ```
 
-The Agent echoes the `messageId` in the response (or assigns one if the client didn't provide it):
+The Agent echoes this as `userMessageId` in the response (or assigns one if the client didn't provide it):
 
 ```json
 {
   "jsonrpc": "2.0",
   "id": 2,
   "result": {
-    "messageId": "4c12d49b-729c-4086-bfed-5b82e9a53400",
+    "userMessageId": "4c12d49b-729c-4086-bfed-5b82e9a53400",
     "stopReason": "end_turn"
   }
 }
@@ -144,13 +144,25 @@ If the Agent sends `user_message_chunk` updates (e.g., during `session/load`), i
 
 The `messageId` field would be:
 
-- **Optional** on `agent_message_chunk`, `user_message_chunk`, and `agent_thought_chunk` updates
-- **Optional** in `session/prompt` requests (client-provided) and responses (agent-provided or echoed)
-- **UUID format** - Both clients and agents SHOULD use UUID format to avoid collisions
+- **Optional** on `agent_message_chunk`, `user_message_chunk`, `agent_thought_chunk` updates and `session/prompt` requests (as `messageId`)
+- **Optional** in `session/prompt` responses (as `userMessageId`)
+- **UUID format** - Both clients and agents MUST use UUID format to ensure collision avoidance
 - **Unique per message** within a session
 - **Stable across chunks** - all chunks belonging to the same message share the same `messageId`
 - **Opaque** - Implementations treat it as an identifier without parsing its structure
 - **Dual-origin** - Clients generate IDs for user messages, agents generate IDs for agent messages
+
+### Message ID Acknowledgment
+
+When a Client includes `messageId` in a `session/prompt` request, the Agent's response behavior indicates whether message IDs are supported:
+
+- **If the Agent supports message IDs** and records the client-provided ID, it MUST include `userMessageId` in the response with the same value.
+- **If the Agent assigns its own ID** (when the client didn't provide one), it SHOULD include `userMessageId` in the response so the client knows the assigned ID.
+- **If the Agent does not support message IDs** or chooses not to record it, it MUST omit `userMessageId` from the response.
+
+Clients MUST interpret the absence of `userMessageId` in the response as confirmation that the message ID was **not recorded**. Clients SHOULD NOT rely on that ID for subsequent operations (e.g., message editing, truncation, or forking) when `userMessageId` is absent.
+
+This creates predictable behavior: the presence of `userMessageId` in the response is an explicit acknowledgment that the Agent recorded the ID and will recognize it in future operations.
 
 ## Shiny future
 
@@ -200,22 +212,22 @@ Example future editing capability:
 1. **Update schema** (`schema/schema.json`):
    - Add optional `messageId` field (type: `string`) to `ContentChunk` (used by `AgentMessageChunk`, `UserMessageChunk`, `AgentThoughtChunk`)
    - Add optional `messageId` field (type: `string`) to `PromptRequest` (client-provided)
-   - Add optional `messageId` field (type: `string`) to `PromptResponse` (agent-provided or echoed)
+   - Add optional `userMessageId` field (type: `string`) to `PromptResponse` (echoed or agent-assigned)
 
 2. **Update Rust SDK** (`rust/client.rs` and `rust/agent.rs`):
    - Add `message_id: Option<String>` field to `ContentChunk` struct
    - Add `message_id: Option<String>` field to `PromptRequest` struct
-   - Add `message_id: Option<String>` field to `PromptResponse` struct
-   - Update serialization to include `messageId` in JSON output when present
+   - Add `user_message_id: Option<String>` field to `PromptResponse` struct
+   - Update serialization to include `messageId`/`userMessageId` in JSON output when present
 
 3. **Update TypeScript SDK** (if applicable):
    - Add `messageId` field to corresponding types
 
 4. **Update documentation** (`docs/protocol/prompt-turn.mdx`):
-   - Document the `messageId` field and its semantics
-   - Clarify that clients can provide `messageId` for user messages
-   - Clarify that agents generate `messageId` for agent messages
-   - Show that `messageId` can be provided in prompt requests and is returned in responses
+   - Document the `messageId` and `userMessageId` fields and their semantics
+   - Clarify that clients can provide `messageId` for user messages in prompt requests
+   - Clarify that agents echo the ID as `userMessageId` in prompt responses
+   - Clarify that agents generate `messageId` for agent messages in chunks
    - Add examples showing message boundaries
    - Explain that `messageId` changes indicate new messages
 
@@ -273,7 +285,7 @@ The proposed approach with `messageId` is:
 
 **Both clients and agents can generate message IDs**, each for their own messages:
 
-- **For user messages**: The Client generates a UUID and includes it as `messageId` in the `session/prompt` request. The Agent echoes this ID in the response. If the client doesn't provide one, the agent MAY assign one.
+- **For user messages**: The Client generates a UUID and includes it as `messageId` in the `session/prompt` request. The Agent echoes this ID as `userMessageId` in the response to confirm it was recorded. If the client doesn't provide one, the Agent MAY assign one and return it in the response so the client knows the assigned ID.
 - **For agent messages**: The Agent generates the UUID when creating its response and includes it in session update chunks.
 
 This differs from other protocol identifiers (`sessionId`, `terminalId`, `toolCallId`) which are agent-generated, but provides practical benefits:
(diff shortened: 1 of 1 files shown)


docs/rfds/message-id.mdx
```diff
@@ -0,0 +1,334 @@
...
+- **Simple** - Just one new field with clear semantics
+- **Flexible** - Enables future capabilities without further protocol changes
+- **Consistent** - Aligns with how other resources (sessions, terminals, tool calls) are identified in the protocol
+- **Centralized** - Agent as single source of truth for all IDs simplifies uniqueness guarantees
+
+### Who generates message IDs?
+
+The **Agent generates all message IDs**, for both user and agent messages:
+
+- **For user messages**: When the Client sends `session/prompt`, the Agent assigns a message ID and returns it as `messageId` in the response
```
anna239: But a response to the `session/prompt` comes only after all `session\update` messages are streamed https://agentclientprotocol.com/protocol/prompt-turn . We also plan to support user messages in the middle of the agent's interactions, in this case it's also not clear how to return messageId to the client. 
I like your proposal a lot, but this particular part about server generating client messages' ids looks a little cumbersome to me

docs/rfds/message-id.mdx
```diff
@@ -0,0 +1,334 @@
...
+// Tool call happens
+{ "sessionUpdate": "tool_call", ... }
+
+// Another agent message - definitely a new message
+{ "sessionUpdate": "agent_message_chunk", "content": { "type": "text", "text": "Fixed the issues." } }
+```
+
+## What we propose to do about it
+
+Add a `messageId` field to `AgentMessageChunk` and `UserMessageChunk` session updates, and to the `session/prompt` response. This field would:
```
benbrandt: I agree with @anna239 here... I am unclear on why a session/prompt response would need a messageId?

If anything that would feel more like a "turnId" which I'm not sure if that would always hold up either.

Basically a prompt would create one or more "messages" , which would happen via the chunks, and then I don't think we need anything on the prompt, unless I am missing something.

We could introduce a new `session/update` type of "message start" and "message end" if needed to carry extra metadata... but I am not sure how useful that is and would hold off unless we have a strong reason to do it.

I think as proposed this ultimately has similar behavior to now (you have to watch the chunks for a new message id) but does allow you to split messages of the same type, which is useful.

And it creates a nice anchor for other features as you mentioned

docs/rfds/message-id.mdx
```diff
@@ -0,0 +1,334 @@
...
+```
+
+The `messageId` field would be:
+
+- **Required** on `agent_message_chunk` and `user_message_chunk` updates
+- **Required** in `session/prompt` responses as `messageId`
+- **Unique per message** within a session
+- **Stable across chunks** - all chunks belonging to the same message share the same `messageId`
+- **Opaque** - Clients treat it as an identifier without parsing its structure
+- **Agent-generated** - The Agent generates and manages all message IDs, consistent with how the protocol handles `sessionId`, `terminalId`, and `toolCallId`
```
benbrandt: How do we handle messageIds for user messages? Are they generated by the client? 
The agent is also allowed to supply user messages, it is true, but I feel we need to also allow the user messages from the prompt to have an id as well

If we allow both sides to generate though, we may need to require something like a uuid or have a prefix or something to avoid collisions

docs/rfds/message-id.mdx
```diff
@@ -0,0 +1,334 @@
...
+  }
+}
+```
+
+## Implementation details and plan
+
+### Phase 1: Core Protocol Changes
+
+1. **Update schema** (`schema/schema.json`):
+   - Add required `messageId` field (type: `string`) to `AgentMessageChunk`
```
benbrandt: To make this a non-breaking change, we will need to have this be `optional` for now.
We can add `SHOULD` language to the spec, but changing that to a `MUST`/`required` will need to be a v2 change

docs/rfds/message-id.mdx
```diff
@@ -0,0 +1,334 @@
...
+   - Use IDs to properly group chunks into messages
+   - Demonstrate clear message boundary rendering
+
+### Backward Compatibility
+
+Since this adds a **required** field, this would be a **breaking change** and should be part of a major version bump of the protocol. Agents and Clients will need to coordinate upgrades.
+
+Alternatively, the field could initially be made **optional** to allow gradual adoption:
+
+- Agents that support it advertise a capability flag during initialization
```
benbrandt: I think we can do optional, and then features that require a messageId would have an implicit flag in the sense that those methods would _require_ a messageId and you wouldn't have one

docs/rfds/message-id.mdx
benbrandt: Agreed. An alternate approach could be to introduce a new session/update that assigns it... But I wonder if we just need a way to get unique ids from both sides.

docs/rfds/message-id.mdx
```diff
@@ -0,0 +1,334 @@
...
+
+The protocol doesn't require message IDs to be stable across session loads, though Agents MAY choose to make them stable if their implementation supports it.
+
+### Does this apply to other session updates like tool calls or plan updates?
+
+This RFD specifically addresses `agent_message_chunk` and `user_message_chunk` updates. Other session update types (like `tool_call`, `agent_thought_chunk`, `plan`) already have their own identification mechanisms:
+
+- Tool calls use `toolCallId`
+- Plan entries can be tracked by their position in the `entries` array
+- Agent thoughts could benefit from message IDs if they're considered distinct messages
```
benbrandt: What is the reason not to add it to thoughts now? Wouldn't you also want some distinction there?

And could one message contain two tool calls? 

I guess I am realizing it would make sense to better define what a "message" is in the protocol, and that might better guide which things need or don't need a message id

benbrandt (commented):
Nice start! I have a few comments, but I think this is definitely a much needed feature

docs/rfds/message-id.mdx
anna239: Strictly speaking, you want them to be unique only among client\server messages. A message role (user/assistant) can be treated as a part of the id, but this complicates things in the scenario with multiple clients. But probably we can be safe just having UUIDs there, wdyt?

docs/rfds/message-id.mdx
SteffenDE: When thinking about session/fork from a specific message, I think returning a messageId for that turn in the prompt response is necessary. I don't think the client should provide IDs, since agents might not support that, so we'd end up with a similar problem that wrappers like claude-code-acp would need to store state themselves.

benbrandt: Yeah I've been wondering about this one as well... but I'm not really sure how to best get a message id _back_ for the prompt message

benbrandt: We can use 
```
  --replay-user-messages                            Re-emit user messages from stdin back on stdout for acknowledgment (only works with --input-format=stream-json and --output-format=stream-json)
  ```
  to potentially get the info needed actually. But yeah, we'd need some message on ACP to set this... this could tie into a revamp of the session updates (early idea here: https://agentclientprotocol.zulipchat.com/#narrow/channel/543470-rfds/topic/Pre-RFD.3A.20Changing.20the.20prompt.2Fnew.20and.20session.2Fupdate.20interact/with/561997764)

I wonder if there is a two-stage option here, where we start adding ids for agent messages (optionally) and then come up with a way for the agent to report back user ids.

Because if we add more methods on the session like truncation or edit or fork etc, we need stable ids the agent can reason about for sure

File: docs/rfds/message-id.mdx (hunk 3 above)
Comment (Nemtecl) [reply docs/rfds/message-id.mdx]: Answering in [this thread](https://github.com/agentclientprotocol/agent-client-protocol/pull/244/changes#r2537030562) !

File: docs/rfds/message-id.mdx (hunk 2 above)
Comment (Nemtecl) [reply docs/rfds/message-id.mdx]: Continuing in [this thread](https://github.com/agentclientprotocol/agent-client-protocol/pull/244/changes#r2537030562) !

File: docs/rfds/message-id.mdx (hunk 1 above)
Comment (Nemtecl) [reply docs/rfds/message-id.mdx]: Thanks for the great discussion, everyone!

Using a `session/update` mechanism like `user_message_chunk` to echo messages with IDs would create a mapping challenge: how would clients reliably link IDs to their original messages? Content matching isn’t foolproof, so PromptResponse avoids this issue entirely

We could let the client set a temporary `messageId`, then have the agent update it via `session/update` with a permanent one. But is everyone comfortable with the client assigning any ID, even temporarily? No strong opinion on my side

For mid-turn user messages, we can address ID assignment during that feature’s design. For now, let’s focus on what’s concrete and avoid over-engineering for hypotheticals 🙏 ?

Open to all suggestions and compromises to keep this RFD progressing! 🚀

File: docs/rfds/message-id.mdx (hunk 6 above)
Comment (Nemtecl) [reply docs/rfds/message-id.mdx]: Fixed in 2a710305665

File: docs/rfds/message-id.mdx (hunk 5 above)
Comment (Nemtecl) [reply docs/rfds/message-id.mdx]: Fixed in 2a710305665

File: docs/rfds/message-id.mdx (hunk 4 above)
Comment (Nemtecl) [reply docs/rfds/message-id.mdx]: Fixed in 2a710305665

**Comment by benbrandt** [comment]
@Nemtecl @anna239 been thinking about this... What do you all think about _only_ supplying message ids for agent messages for now?

Would this solve enough problems without creating new ones?
I think getting ids for user messages is going to be a bit tricky... but I wonder if we can make progress by at least introducing message ids for the agent side?

**Comment by denofevil** [comment]
@benbrandt by looking at the elevator pitch, we need to identify client messages, not just agent ones:
> lays the groundwork for future capabilities like message editing.

We're looking also at `messageId` for the purpose of deduplicating messages on `session/load` (and echoing client prompt to multiple clients, I'll be drafting that shortly). I suppose we can kill two birds with one stone here if we agree on the id generation scheme, e.g. both sides could generate them as UUID.

**Comment by benbrandt** [comment]
@denofevil the issue I have is that not all agents (such as codex) allow for passing user message ids in. 
It's possible we can hope for the best, but would need some work to make sure this is reliable enough for all of the operations you want to do

**Comment by denofevil** [comment]
I might be not seeing the whole picture here. AFAIU codex goes through an adapter, adapter can just don't pass messageId to the agent. If we're speaking about editing the message, I suppose that would be under an agent's capability anyway

**Comment by Nemtecl** [comment]
Hi @benbrandt and @denofevil 👋 

I agree with @denofevil’s perspective. This will not only enable message identification but also unlock other features, such as edit functionality

For agents that don’t yet support messageId, we could introduce an agent capability flag for feature that need message id. Clients could then fall back to alternative workflows, such as fork sessions, to maintain compatibility while encouraging adoption of the new standard

As a first step, I’m open to making the messageId configurable, either generated by the client or the agent, using a UUID format. This flexibility should help ease the transition and accommodate different implementations

Let me know if this aligns with your vision, and I’m happy to address the changes !

**Comment by benbrandt** [comment]
@Nemtecl right but I guess you can also implement edit (this is how we do it in zed and codex seems to do something similar) by truncating the message history to before the user message (basically an inclusive index of the agent message id) and then submit the new user message with the edited text. 

So I feel like we could still support a lot of these flows with just agent message ids. The replay aspect @denofevil brings up makes sense though.

I am fine with aligning on uuids. Regarding capabilities, this makes sense. This feels like something we could make mandatory in a major version bump, but we'll need to make opt-in for now.

If you dont' mind doing one more round of updates to the RFD, we can likely move to draft stage and see what issues we run into with implementation

**Comment by Nemtecl** [comment]
Hi @benbrandt 👋 

Made the changes in 9d6bc2aafcaee3cc1cb8c833602295b2a9ab7667 !

File: docs/rfds/message-id.mdx [hunk 7]
```diff
@@ -0,0 +1,324 @@
...
+```
+
+The Agent echoes the `messageId` in the response (or assigns one if the client didn't provide it):
+
+```json
+{
+  "jsonrpc": "2.0",
+  "id": 2,
+  "result": {
+    "messageId": "4c12d49b-729c-4086-bfed-5b82e9a53400",
```
benbrandt: Hmm I see why this is here (associate the turn it was in response to) but feels weird here as just `messageId`
Maybe solvable with documentation of what this means, as in that this is the result for the prompt turn provided by the corresponding user message id. But that is also what the `id` field above is for.

docs/rfds/message-id.mdx
```diff
@@ -0,0 +1,324 @@
...
+```
+
+The `messageId` field would be:
+
+- **Optional** on `agent_message_chunk`, `user_message_chunk`, and `agent_thought_chunk` updates
+- **Optional** in `session/prompt` requests (client-provided) and responses (agent-provided or echoed)
+- **UUID format** - Both clients and agents SHOULD use UUID format to avoid collisions
+- **Unique per message** within a session
+- **Stable across chunks** - all chunks belonging to the same message share the same `messageId`
+- **Opaque** - Implementations treat it as an identifier without parsing its structure
```
benbrandt, line 152: Just a thought, not a blocker: 
- we say it should be a UUID, but also it should be opaque. However, because both sides can create/emit them, I think we may need to say it MUST be a uuid and the other side can rely on it being so. 

What do you think?

docs/rfds/message-id.mdx
```diff
@@ -0,0 +1,324 @@
...
+When the Client sends a user message via `session/prompt`, it can optionally include a `messageId`:
+
+```json
+{
+  "jsonrpc": "2.0",
+  "id": 2,
+  "method": "session/prompt",
+  "params": {
+    "sessionId": "sess_abc123def456",
+    "messageId": "4c12d49b-729c-4086-bfed-5b82e9a53400",
```
benbrandt, line 81: One potential issue I see with this:

- Agents _implicitly_ express their support of message ids by including it in their messages
- However, if a client sends an id, there is an _implicit_ assumption that the agent recorded it. For other features to work around the client-sent messages, they have to assume that that message id was respected
- We can kind of implicitly understand the agent did something with the id by receiving the message id in the response. That would be one reason to leave it there, but I wonder if there is a better structure or name for that field.

docs/rfds/message-id.mdx
benbrandt: Also, if the user did not pass this in, would the agent respond with the message id the agent assigned to that message?

docs/rfds/message-id.mdx
Nemtecl: Would you like to have some agent and client capabilities for `messageId` ?

Another solution could be that, until `messageId` are mandatory, future features that may rely around `messageId` would be flagged in capabilties

For example 

```ts
  agentCapabilities: {
    // ...
    editMessage: "from-id" | boolean,
  }
```

Open to any suggestion on how to address the issue you mentioned !

Nemtecl:
Hi @benbrandt 

Just made some update in https://github.com/agentclientprotocol/agent-client-protocol/pull/244/commits/d5c170fb8a767a1907d51fec7624dd4937fa8a46

Hopefully this should tackle your comments 🙏

benbrandt:
Thanks for going several rounds on this! 
I think this is good enough to start implementation on. We all know we want it, we still have some questions, but some questions that will be easier answered by building.

Implementation can begin. I had to push this to another branch for some reason, but it is here: https://github.com/agentclientprotocol/agent-client-protocol/pull/536

denofevil:
@Nemtecl @benbrandt is any of you working on adding this into the unstable schema? We'd like to play with this in regards to #533. If no one objects, we can contribute an update to the schema

SteffenDE:
@denofevil see https://github.com/agentclientprotocol/agent-client-protocol/pull/581.

benbrandt:
yes! I am wrapping up some things today/tomorrow to package as a schema release, will get this in there and get it out soon!

PR #248, opened 2025-11-18
RFD for Per-request cancellation mechanism

nerzhulart:
RDF for Per-request cancellation mechanism

The draft was started here https://github.com/agentclientprotocol/agent-client-protocol/pull/183, but here is RFD for this feature.

### DIFF
--- a/docs/rfds/request-cancellation.mdx
+++ b/docs/rfds/request-cancellation.mdx
@@ -2,8 +2,8 @@
 title: "Request Cancellation Mechanism"
 ---
 
-Author(s): [Artem Bukhonov](https://github.com/nerzhulart)
-Champion: [@benbrandt](https://github.com/benbrandt)
+- Author(s): [Artem Bukhonov](https://github.com/nerzhulart)
+- Champion: [@benbrandt](https://github.com/benbrandt)
 
 ## Elevator pitch
 
--- a/docs/updates.mdx
+++ b/docs/updates.mdx
@@ -5,23 +5,30 @@ rss: true
 ---
 
 <Update label="November 20, 2025" tags={["RFD"]}>
+## $/cancelRequest RFD moves to Draft stage
+
+The RFD for adding a "$/cancelRequest" method to the protocol has been moved to Draft stage. Please review the [RFD](./rfds/request-cancellation) for more information on the current proposal and provide feedback as work on the implementation begins.
+
+</Update>
+
+<Update label="November 20, 2025" tags={["RFD"]}>
 ## session/fork RFD moves to Draft stage
 
-The RFD for adding a "session/fork" method to the protocol has been moved to Draft stage. Please review the [RFD](https://agentclientprotocol.com/rfds/session-fork) for more information on the current proposal and provide feedback as work on the implementation begins.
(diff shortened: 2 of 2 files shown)


docs/rfds/request-cancellation.mdx
```diff
@@ -0,0 +1,258 @@
...
+
+Once implemented, this enables:
+
+- **SDK integration layer**: Default mechanism that ACP SDKs can automatically wire to native language cancellation (C# CancellationToken, Kotlin Job, Go context.Context, JavaScript AbortController, etc.)
+- Individual JSON-RPC request cancellation without affecting other concurrent requests
+- Universal fallback for any request when feature-specific cancellation methods don't exist
+- Consistent cancellation behavior from both external `$/cancelRequest` and internal cancellation triggers
+- Standard error response (`-32800`) or partial results when requests are cancelled regardless of cancellation source
+
+The mechanism complements existing `session/cancel` for prompt turns while providing granular per-request control.
```
benbrandt, line 45: The more I think about it, the more I think we could use this for session/cancel as well, and deprecate/remove it in v2

docs/rfds/per-request-cancellation.mdx
```diff
@@ -0,0 +1,258 @@
...
+
+### Protocol Changes
+
+#### Capability Declaration
+
+Cancellation support must be declared during the capabilities exchange. Add new optional capability fields:
+
+```typescript
+interface CancellationCapabilities {
+  supportsRequestCancellation?: boolean; // Default: false
```
benbrandt: I wonder if this could be an empty object for now? And it assumes if the object is there that you support general request cancellation? And then we can add more granular knobs later?

Or, we change this to just `request: boolean`

docs/rfds/request-cancellation.mdx
```diff
@@ -0,0 +1,258 @@
...
+
+- Check the recipient's capabilities before sending `$/cancelRequest`
+- Only send `$/cancelRequest` if the recipient declared `supportsRequestCancellation: true`
+- Not send `$/cancelRequest` notifications to parties that don't support them
+
+The **receiving party** is **NOT** required to:
+
+- Perform special handling for unsupported cancellation requests
+- Return custom errors for unsupported `$/cancelRequest` notifications
+- The standard JSON-RPC "method not found" mechanism will handle unsupported requests
```
benbrandt: Since this is a notification, which response is this referring to?

docs/rfds/request-cancellation.mdx
```diff
@@ -0,0 +1,258 @@
...
+### How does this relate to existing cancellation mechanisms like `session/cancel`?
+
+The `$/cancelRequest` mechanism is complementary to feature-specific cancellation:
+
+- `$/cancelRequest`: Generic cancellation for any JSON-RPC request by ID
+- `session/cancel`: Feature-specific cancellation with additional semantics (e.g., cancels entire prompt turn context, triggers specific cleanup logic)
+
+Both mechanisms serve different purposes:
+
+**Feature-specific methods** like `session/cancel` provide:
```
benbrandt, line 160: I think as long as we allow for a cancellation notification to do some amount of cleanup, we could still model this feature specific ones the same way

PR #262, opened 2025-11-24
Added aizen acp client

wiedymi:
I am a developer of Aizen. I would like to add my acp client to the clients section in the docs.


wiedymi:
Thank you!

PR #279, opened 2025-11-28
add separate resumeToken for session/resume

SteffenDE:
To make implementing resume easy for Claude Code, I'm proposing session/resume to use a separate `resumeToken`.

Because of the way the Claude SDK works, `claude-code-acp` only gets the internal Claude Session ID after sending the first prompt. Therefore the ACP `sessionId` is completely separate from Claude's session id. Without this, `claude-code-acp` would need to keep a separate session mapping somewhere on disk to implement resuming.

### DIFF
--- a/docs/rfds/session-resume.mdx
+++ b/docs/rfds/session-resume.mdx
@@ -46,6 +46,48 @@ the stored messages.
 
 > Tell me more about your implementation. What is your detailed implementation plan?
 
+We propose to add a new "session/resume" method. Agents must declare this option is
+available by including `sessionCapabilities: { resume : {} }` in its capabilities.
+The object is reserved to declare future capabilities.
+
+To make implementation in existing and future agents easier, we propose to have a separate
+token for resuming, which the agent can use to map to an internal session:
+
+```json
+{
+  "jsonrpc": "2.0",
+  "id": 1,
+  "method": "session/resume",
+  "params": {
+    "resumeToken": "opaque-token"
+  }
+}
+```
+
+And the agent would respond with optional data such as config options, the same as `session/load`.
+
+When implementing `session/resume`, the agent sends a `session/update` notification with the token:
+
+```json
+{
+  "jsonrpc": "2.0",
+  "method": "session/update",
+  "params": {
+    "sessionId": "sess_123",
+    "update": {
+      "sessionUpdate": "resumeToken",
+      "resumeToken": "opaque-token"
+    },
+  }
+}
+```
+
+Agents may update the `resumptionToken` at any time if necessary. Clients should use the latest
+token received to continue the chat from where they left off.
+
+Agents may only support resuming chats in some cases. Therefore, an agent may not send a `resumeToken`,
+even if it declares the capability.
+
 ## Frequently asked questions
 
 > What questions have arisen over the course of authoring this document or during subsequent discussions?


docs/rfds/session-resume.mdx
```diff
@@ -46,6 +46,48 @@ the stored messages.
 
 > Tell me more about your implementation. What is your detailed implementation plan?
 
+We propose to add a new "session/resume" method. Agents must declare this option is
+available by including `sessionCapabilities: { resume : {} }` in its capabilities.
```
SteffenDE, line 50: I already used the new `sessionCapabilities` proposed by `session/list`. Happy to change that to a top level key.

docs/rfds/session-resume.mdx
```diff
@@ -46,6 +46,48 @@ the stored messages.
...
+    },
+  }
+}
+```
+
+Agents may update the `resumptionToken` at any time if necessary. Clients should use the latest
+token received to continue the chat from where they left off.
+
+Agents may only support resuming chats in some cases. Therefore, an agent may not send a `resumeToken`,
+even if it declares the capability.
```
benbrandt, line 89: The nice thing about using a `token` is that it would allow for us to have multiple resume points in the conversation i.e. for forking

But I guess this is somewhat different in that resume kind of always assumes resuming from whatever point the agent decides correct? Like the last message?

So, I guess in general i like the idea.. but I also feel like we are applying a band-aid at the protocol level for lack of support from the Claude Code SDK...

Because I guess what isn't clear to me is you would get this session/update shortly after starting the conversation and then it is somewhat unclear where you are actually resuming from... 
And should this resume token be somehow returned by the `session/list` endpoint, if supported?

SteffenDE: > But I guess this is somewhat different in that resume kind of always assumes resuming from whatever point the agent decides correct? Like the last message?

Yes. I'd also say the main difference between resume and fork (with message id) is that fork assumes an already running session, while resume does not.

> but I also feel like we are applying a band-aid at the protocol level for lack of support from the Claude Code SDK...

It would be nice if we don't need it, yes. The token approach feels very flexible though and as long as we document that resuming should always resume from the most recent interaction, I think we're good. "Resuming" from a different point in time is something I'd only consider for fork.

The "getting it shortly after starting the conversation" is somewhat similar to agent modes though, since those also don't change most of the time.

> And should this resume token be somehow returned by the session/list endpoint, if supported?

I'd say so, yes!

benbrandt: So the resume token is just an indication that this session is now resumable?
vs just being able to call resume on a session?

benbrandt: Like i guess, if it is always from the most recent interaction, why wouldn't I just ask to resume a given session?

SteffenDE: > So the resume token is just an indication that this session is now resumable?

Correct. This PR proposes that if you don't get a resume token, you cannot resume that session.

> Like i guess, if it is always from the most recent interaction, why wouldn't I just ask to resume a given session?

Right, the whole point of the PR is to make it more flexible for the agents implementing resume. If you only send the ACP session ID for resuming, Claude wouldn't know what to do, as it has no mapping of ACP sessions to Claude sessions. The resume token bridges that gap without requiring the agent to keep persistent state. But I can understand if you don't want to go that route. The alternative is that claude-code-acp stores state somewhere on disk, e.g. in `$HOME/.claude-code-acp`.

benbrandt: I guess my current thinking is it would be helpful to see more agent implementations to see if this extra token is helpful or not .. or if it is just because Claude Code doesn't allow us to get an id in time.

benbrandt:
@SteffenDE can we close this one now that we solved the session id issue in claude code?
And revisit as we implement in others?

SteffenDE:
Agree!

PR #288, opened 2025-11-28
docs(rfd): Draft: Session info update notification

ignatov:
- Adds session_info_update variant to SessionUpdate
- Enables real-time session metadata updates (title, _meta)
- Follows existing notification pattern (available_commands_update, current_mode_update)
- All fields optional for partial updates
- Must stay aligned with SessionInfo structure

### DIFF
--- a/docs/rfds/session-info-update.mdx
+++ b/docs/rfds/session-info-update.mdx
@@ -3,7 +3,6 @@ title: "Session Info Update"
 ---
 
 - Author(s): [@ignatov](https://github.com/ignatov)
-- Champion: [@benbrandt](https://github.com/benbrandt)
 
 ## Elevator pitch
 
--- a/docs/docs.json
+++ b/docs/docs.json
@@ -97,7 +97,8 @@
               "rfds/session-fork",
               "rfds/request-cancellation",
               "rfds/session-resume",
-              "rfds/meta-propagation"
+              "rfds/meta-propagation",
+              "rfds/session-info-update"
             ]
           },
           { "group": "Preview", "pages": [] },
--- a/docs/updates.mdx
+++ b/docs/updates.mdx
@@ -5,6 +5,13 @@ rss: true
 ---
 
 <Update label="December 3, 2025" tags={["RFD"]}>
+## session_info_update notification RFD moves to Draft stage
+
+The RFD for adding a new `session_info_update` variant on the `session/update` notification in the protocol has been moved to Draft stage. Please review the [RFD](./rfds/session-info-update) for more information on the current proposal and provide feedback as work on the implementation begins.
+
+</Update>
+
+<Update label="December 3, 2025" tags={["RFD"]}>
 ## _meta Propagation RFD moves to Draft stage
 
 The RFD for providing more guidance on how the `_meta` parameter should be used within the protocol has been moved to Draft stage. Please review the [RFD](./rfds/meta-propagation) for more information on the current proposal and provide feedback as work on the implementation begins.


ignatov:
my personal apologies, it's very AI generated

docs/rfds/session-info-update.mdx
```diff
@@ -0,0 +1,372 @@
...
+
+3. **Consistent protocol patterns** - Session metadata updates work like other dynamic session properties (commands, modes), creating a unified model
+
+4. **Bidirectional workflows** - Combined with a potential future request method:
+   - User renames session → client sends request → agent acknowledges with `session_info_update` notification
+   - Agent auto-generates title → sends `session_info_update` notification → client displays it
+
+5. **Enhanced use cases**:
+   - Session templates that auto-set titles and tags
+   - Progress indicators via `_meta`
```
benbrandt, line 162: This one is huge. i see people trying to send random empty content updates for things like this, and this would help a lot 👍🏻

docs/rfds/session-info-update.mdx
```diff
@@ -0,0 +1,372 @@
...
+
+- `sessionId` is already in the notification params
+- `cwd` is immutable (set in `session/new`, never changes)
+- Keeps update focused on mutable metadata
+
+**How do `_meta` updates work?**
+
+- **Merge semantics**: New `_meta` fields are merged with existing ones
+- To clear a specific field: `"_meta": { "fieldName": null }`
+- To clear all custom metadata: `"_meta": null`
```
benbrandt, line 283: This is a new one that we might have to handle carefully
Right now _meta has always been a black box, so if we want this behavior, we may need to apply this as a general recommendation for things that are updated multiple times (like messages once we have message ids, or tool calls)

We could sketch this behavior out for these updates, add a **SHOULD** in the other spots and put it on the list for v2, where we may also want to provide more guidance like MCP does on the _meta field names

docs/rfds/session-info-update.mdx
```diff
@@ -0,0 +1,372 @@
...
+The updated metadata should persist and be reflected in subsequent `session/list` calls. The notification provides real-time updates to connected clients, while `session/list` provides the current state for discovery.
+
+### Can clients trigger title updates?
+
+This RFD covers agent-initiated updates. Client-initiated updates could work by:
+
+1. Client sends a prompt asking to rename session
+2. Agent updates its internal state
+3. Agent sends `session_info_update` notification
+4. Client receives and displays the update
```
benbrandt, line 316: I'm not sure this would always work, I might remove this and just state that client-updates would be handled in a future RFD

docs/rfds/session-info-update.mdx
```diff
@@ -0,0 +1,372 @@
...
+
+A future RFD could add explicit request/response for this if needed.
+
+### What if multiple updates are sent in quick succession?
+
+Clients should apply updates incrementally in order. Each notification represents a delta, not a full replacement (except for fields explicitly set to `null`).
```
benbrandt, line 322: We don't currently have this unset behavior in tool calls (probably should, but would be a different topic)
Just pointing out we would differ here, but could plan to align these in the future

docs/rfds/session-info-update.mdx
```diff
@@ -0,0 +1,372 @@
...
+
+### Should `updatedAt` be automatically set by the agent?
+
+Yes, typically the agent should update this timestamp when any session activity occurs, not just when metadata changes. However, including it in `session_info_update` allows agents to explicitly control it if needed.
```
benbrandt, line 326: It is an interesting question from the client point of view, where likely updatedAt would be updated fairly constantly.
However, the agent could decide when to update that timestamp as you say, and the next time the list is refreshed it would also be updated. And it is likely less important when the session is loaded like the title, as you can see when the last message was potentially.

So it isn't required that the agent sends an update whenever the agent would update this time, but they _can_ if they feel like it's important to communicate eagerly

docs/rfds/session-info-update.mdx
```diff
@@ -0,0 +1,372 @@
+---
+title: "Session Info Update"
+---
+
+- Author(s): [@ignatov](https://github.com/ignatov)
+- Champion: [@benbrandt](https://github.com/benbrandt)
```
benbrandt: Lucky you, core maintainer, you don't need a champion :D though I am happy to help

benbrandt (approved):
Overall, I like it, have a few questions, but will likely do a cleanup pass and merge regardless and we can update the RFD as we work on the draft implementation.

docs/rfds/session-info-update.mdx
ignatov: okay, let's omit _meta for now, my aim is to add session_info update and involve it further

docs/rfds/session-info-update.mdx
ignatov: okay

docs/rfds/session-info-update.mdx
benbrandt: I guess there are two options here: 
- omit _meta
- decide how/if we want to handle merging _meta

benbrandt: I actually would say we _shouldn't_ omit _meta because we always offer it. I just realized we don't have _meta well-defined enough to maybe adequately require merging... though this is maybe possible

docs/rfds/session-info-update.mdx
ignatov: I'll be glad to receive your help and get a guidance on how to transform a proposal to changes in rust code

PR #289, opened 2025-11-28
Add ACP Agent Registry RFD

ignatov:
- Define canonical manifest format (agent.json) for ACP agents
- Specify registry structure with <id>/ folders
- Add distribution section for binary targets across platforms
- Include icon support for light/dark themes
- Outline implementation phases and aggregation tooling

### DIFF
--- a/docs/docs.json
+++ b/docs/docs.json
@@ -113,7 +113,8 @@
               "rfds/session-info-update",
               "rfds/agent-telemetry-export",
               "rfds/proxy-chains",
-              "rfds/session-usage"
+              "rfds/session-usage",
+              "rfds/acp-agent-registry"
             ]
           },
           { "group": "Preview", "pages": [] },
--- a/docs/updates.mdx
+++ b/docs/updates.mdx
@@ -5,6 +5,13 @@ rss: true
 ---
 
 <Update label="January 1, 2026" tags={["RFD"]}>
+## Agent Registry RFD moves to Draft stage
+
+The RFD for creating an Agent Registry has been moved to Draft stage. Please review the [RFD](./rfds/acp-agent-registry) for more information on the current proposal and provide feedback as work on the implementation begins.
+
+</Update>
+
+<Update label="January 1, 2026" tags={["RFD"]}>
 ## Session Usage RFD moves to Draft stage
 
 The RFD for adding a new `usage_update` variant on the `session/update` notification and `usage` field on prompt responses in the protocol has been moved to Draft stage. Please review the [RFD](./rfds/session-usage) for more information on the current proposal and provide feedback as work on the implementation begins.


docs/rfds/acp-agent-registry.mdx
```diff
@@ -0,0 +1,125 @@
...
+   - CI: validate manifests, enforce slug uniqueness, check asset sizes, generate aggregate artifacts.
+3. **Aggregated outputs**:
+   - `registry.json`: deterministic list of all agents with JSONC stripped.
+4. **Distribution & search**:
+   - Clients fetch `registry.json` from a pinned release or `https://agentclientprotocol.com/registry.json` or `https://registry.agentclientprotocol.com`.
+   - Static site offers filters for capability, protocol version, deployment, auth model, and tags.
+
+## Shiny future
+
+- Agent maintainers make PRs to update their manifests; CI keeps data clean.
```
anna239, line 108: how do agent updates work? Agent authors upload new agent.json to repository with updated version?

docs/rfds/acp-agent-registry.mdx
```diff
@@ -0,0 +1,125 @@
...
+      "cmd": "./someagent",
+      "args": ["acp"],
+    },
+    "windows-x86_64": {
+      "archive": "https://github.com/example/someagent/releases/latest/download/someagent-windows-x64.zip",
+      "cmd": "./someagent.exe",
+      "args": ["acp"],
+      "env": {
+        "SOMEAGENT_MODE_KEY": "",
+      },
```
yordis, line 87: Since this is a spec, should we care about knowing if the secret is required or not?

docs/rfds/acp-agent-registry.mdx
```diff
@@ -0,0 +1,125 @@
...
+{
+  "id": "someagent",
+  "name": "SomeAgent",
+  "version": "1.0.0",
+  "schema_version": "1",
+  "description": "Agent for code editing",
+  "homepage": "https://github.com/example/someagent",
+  "repository": "https://github.com/example/someagent",
+  "authors": ["Example Team"],
+  "license": "MIT",
```
yordis, line 53: Should this be an array?

docs/rfds/acp-agent-registry.mdx
```diff
@@ -0,0 +1,125 @@
...
+
+- Let users discover agents directly inside ACP-aware clients.
+- Ensure protocol-version compatibility or capability coverage.
+- Keep metadata consistent (auth requirements, hosting model, license, etc.).
+
+Every editor builds bespoke manifests or scrapes GitHub, leading to duplication and stale data.
+
+## Agent manifest format (core proposal)
+
+Each agent advertises itself via `agent.json` stored under `<id>/` in the registry repo. JSONC keeps things close to ACP’s JSON-centric schemas while remaining human-friendly during authoring. Fields (required unless noted):
```
benbrandt, line 24: I'm curious here... if we are already doing jsonc, which will need some post-processing anyway to serve, would we be better off with something like toml or another format?

Not of a strong opinion here but it seems we have two use cases:
- authoring
- consuming in a client

I can understand the argument that clients probably want to read json, but I guess I just wonder if that also necessitates that to be the authoring language, as long as we provide the tooling in the registry to convert from one to the other

docs/rfds/acp-agent-registry.mdx
benbrandt: SPDX allows for AND and OR already, I think this is fine as a string: https://spdx.github.io/spdx-spec/v2.3/SPDX-license-expressions/

docs/rfds/acp-agent-registry.mdx
```diff
@@ -0,0 +1,125 @@
...
+| `authors`        | Array of author/organization names (mirrors `authors` in the TOML example).                                                                                                                                                                                                                                                                                        |
+| `license`        | SPDX identifier or `"proprietary"`.                                                                                                                                                                                                                                                                                                                                |
+| `capabilities`   | Array of ACP method names implemented (e.g. `["terminal/new","files/read"]`).                                                                                                                                                                                                                                                                                      |
```
benbrandt: Should this be a string? Or just a mirror of the `agentCapabilities` type?

docs/rfds/acp-agent-registry.mdx
```diff
@@ -0,0 +1,125 @@
...
+| `auth`           | Array of auth options for authentication. This is the trickiest part of the schema.                                                                                                                                                                                                                                                                                |
+| `distribution`   | _Optional._ Object mapping target platforms to download/execution info. Each target key follows `<os>-<arch>` format (e.g., `darwin-aarch64`, `linux-x86_64`, `windows-x86_64`). Each target specifies `archive` (download URL), `cmd` (executable path), optional `args` (array of command-line arguments), and optional `env` (object of environment variables). |
```
benbrandt: Why optional? I assume this is largely the point?
Though maybe we offer a way to discriminate between different distribution types in the future?

docs/rfds/acp-agent-registry.mdx
```diff
@@ -0,0 +1,125 @@
...
+| `license`        | SPDX identifier or `"proprietary"`.                                                                                                                                                                                                                                                                                                                                |
+| `capabilities`   | Array of ACP method names implemented (e.g. `["terminal/new","files/read"]`).                                                                                                                                                                                                                                                                                      |
+| `auth`           | Array of auth options for authentication. This is the trickiest part of the schema.                                                                                                                                                                                                                                                                                |
```
anna239: I believe that authorization should happen inside the ACP and be handled through protocol itself, as it can depend on selected model

docs/rfds/acp-agent-registry.mdx
ignatov: yes, we have a GitHub action that builds the unified json with all agents and clients can check it once in a while

docs/rfds/acp-agent-registry.mdx
ignatov: oops :)

docs/rfds/acp-agent-registry.mdx
ignatov: okay, let's discuss that with @benbrandt; he has much more experience here

docs/rfds/acp-agent-registry.mdx
```diff
@@ -0,0 +1,125 @@
...
+| `homepage`       | URL for docs/marketing.                                                                                                                                                                                                                                                                                                                                            |
+| `repository`     | Source repository URL.                                                                                                                                                                                                                                                                                                                                             |
+| `authors`        | Array of author/organization names (mirrors `authors` in the TOML example).                                                                                                                                                                                                                                                                                        |
```
codefromthecrypt: should this refer to the TOML?

Thinking aloud for instructive examples.. I would like to avoid some polarizing patterns where folks in other domains start listing out a subset of contributors who recently touched something.

Maybe "e.g. If GitHub, owner of the repository or if Linux Foundation, name of the most specific project."

docs/rfds/acp-agent-registry.mdx
codefromthecrypt: I assume the value here was intentionally more like a TODO comment, as we'd not commit "trickiest part" as maybe that's too honest for a spec? ;)

docs/rfds/acp-agent-registry.mdx
```diff
@@ -0,0 +1,125 @@
...
+  "repository": "https://github.com/example/someagent",
+  "authors": ["Example Team"],
+  "license": "MIT",
+  "capabilities": ["terminal", "fs/read", "fs/write"],
+  "auth": [
+    {
+      "type": "api_key",
+    },
+  ],
+  "distribution": {
```
codefromthecrypt, line 60: are we expecting the cmd/args to need to vary like this? Is this meant to be used exactly? I guess this is for windows only?

docs/rfds/acp-agent-registry.mdx
codefromthecrypt: food for thought, but I think the distribution manifest could go in many ways.. we may not be able to assume a latest tag for things like this. I ran into this with envoy and here's an example registry where ended up needing to add a date field (though many agents update multiple times a day, envoy never) https://archive.tetratelabs.io/envoy/envoy-versions.json

main concern about the download manifest part is if this digs a hole quite deep. it could be that we make a conventional manifest that is independent from the registry, then the registry assumes that packaing? just ideas.

benbrandt:
@ignatov looking more at the MCP registry, they seem to allow for multiple distribution methods, such as npm, github release, docker, etc

We could still potentially put some cache in front if we want, but I think from an authoring perspective it could be nice to use some similar options?

docs/rfds/acp-agent-registry.mdx
ignatov: let's do just string

docs/rfds/acp-agent-registry.mdx
ignatov: As we still need to do some preprocessing before publishing, if we can easily convert existing extension files from Zed, that may work.

I don’t have many strong opinions here, here's my notes. For clients, choosing JSON is a bit better: fewer dependencies. A second note: it’s more consistent overall. The resulting aggregated output is JSON, the protocol itself is JSON, and the agent descriptor is JSON.

Yes, TOML is growing, but it’s still not a default choice for many ecosystems.

docs/rfds/acp-agent-registry.mdx
ignatov: let's mirror them

docs/rfds/acp-agent-registry.mdx
ignatov: you're

ignatov:
a note on env variables, see https://github.com/augmentcode/auggie-zed-extension/blob/5ee8284c9179a960acc7278f1916647abf7f1de0/extension.toml#L17

for disabling autoupdate, augment passed a flag

Patrick-Erichsen:
Has anyone thought about a broader "ecosystem vision" for the registry, like [what MCP has](https://github.com/modelcontextprotocol/registry/blob/main/docs/design/ecosystem-vision.md)?

I'm curious if [this line](https://github.com/agentclientprotocol/agent-client-protocol/pull/289/changes#diff-86ffb7bd649410a59ed54075d2541a422926f8df7db8ee67e512554e0d1f3288R10) is intentionally ruling out subregistries:

> ACP needs a single, trusted registry...

MCP explicitly [supports subregistries](https://github.com/modelcontextprotocol/registry/blob/main/docs/design/ecosystem-vision.md?plain=1#L22) for curation and extended metadata. I could see that being useful here too.

ignatov:
@Patrick-Erichsen you’re right, I suppose I didn’t express that clearly in the initial PR. The idea is to have a common format, a primary registry, and to allow any number of third-party registries—both private and public ones.

And for sure, we’re looking at MCP as a great example of how to build a thriving ecosystem.

ignatov:
Moreover, it’s a work in progress, and I hope that soon we will show something

codefromthecrypt:
@ignatov please push your latest greatness, rebased etc and let's get the first RFD in! I'm happy to take another pass

ignatov:
@codefromthecrypt here's the current version, it is pretty simple, e.g. for opencode we have the following manifest

```
{
  "id": "opencode",
  "name": "OpenCode",
  "version": "1.0.164",
  "description": "The open source coding agent",
  "repository": "https://github.com/sst/opencode",
  "authors": ["Anomaly"],
  "license": "MIT",
  "icon": "./icon.svg",
  "distribution": {
    "binary": {
      "darwin-aarch64": {
        "archive": "https://github.com/sst/opencode/releases/download/v1.0.164/opencode-darwin-arm64.zip",
        "cmd": "./opencode",
        "args": ["acp"]
      },
      "darwin-x86_64": {
        "archive": "https://github.com/sst/opencode/releases/download/v1.0.164/opencode-darwin-x64.zip",
        "cmd": "./opencode",
        "args": ["acp"]
      },
      "linux-aarch64": {
        "archive": "https://github.com/sst/opencode/releases/download/v1.0.164/opencode-linux-arm64.tar.gz",
        "cmd": "./opencode",
        "args": ["acp"]
      },
      "linux-x86_64": {
        "archive": "https://github.com/sst/opencode/releases/download/v1.0.164/opencode-linux-x64.tar.gz",
        "cmd": "./opencode",
        "args": ["acp"]
      },
      "windows-x86_64": {
        "archive": "https://github.com/sst/opencode/releases/download/v1.0.164/opencode-windows-x64.zip",
        "cmd": "./opencode.exe",
        "args": ["acp"]
      }
    }
  }
}
```

uvx and npx runtime is also supported, I keep the whole registry in private right now, figuring out how to ship agents which don't support auth right now, see #330 

cc @benbrandt and @anna239

ignatov:
Hey!

I published a very early version of the ACP registry: https://github.com/agentclientprotocol/registry

For now, I have listed only agents that have some authentication methods. At the moment, this is mostly opening a browser and starting an HTTP server from the agent process, or launching a terminal via _meta.terminal-auth.

I think that supporting a proper authentication flow (#330) is crucial for smooth distribution in editors and IDEs.

Also, after agreeing on auth flow, it will be cool to speedup the adoption across agents, I think that we can show examples and ask our dear collaborators who're building agents to support the auth flow as soon as possible

docs/rfds/acp-agent-registry.mdx
```diff
@@ -0,0 +1,126 @@
...
+
+| Field            | Description                                                                                                                                                                                                                                                                                                                                                        |
+| ---------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------ |
+| `id`             | Lowercase slug, unique across registry (also the folder name).                                                                                                                                                                                                                                                                                                     |
```
nikomatsakis: I think we should have some "type" field -- the only valid option might be `"agent"` right now, but in the future, it would potentially include `"client"` (that might be too complex) or `"proxy"` (definitely feasible).

docs/rfds/acp-agent-registry.mdx
nikomatsakis: I'd be inclined to remove `auth` for now until things stabilize more.

docs/rfds/acp-agent-registry.mdx
```diff
@@ -0,0 +1,126 @@
...
+| `capabilities`   | Array of ACP method names implemented (e.g. `["terminal/new","files/read"]`).                                                                                                                                                                                                                                                                                      |
+| `auth`           | Array of auth options for authentication. This is the trickiest part of the schema.                                                                                                                                                                                                                                                                                |
+| `distribution`   | Object mapping target platforms to download/execution info. Each target key follows `<os>-<arch>` format (e.g., `darwin-aarch64`, `linux-x86_64`, `windows-x86_64`). Each target specifies `archive` (download URL), `cmd` (executable path), optional `args` (array of command-line arguments), and optional `env` (object of environment variables). |
```
nikomatsakis, line 39: One thing I am wondering: many agents are distributed today as npm packages. This format seems to assume a binary. Granted, the binary might be a standalone npm distro or a shell script that just calls `npx`, but I wonder if we should allow a target key like `npm` to specify an npm package.

ignatov: Please see https://github.com/agentclientprotocol/registry, I pushed a draft of a real schema there. npm and uv packages are supported

docs/rfds/acp-agent-registry.mdx
ignatov: Thanks, I didn't think about that before, can you tell a bit more about proxies? Are they going to be shipped as binaries? npx packages?

codefromthecrypt:
Thx for the example notably I see the issue isn't just .exe but also .zip

Nit lets use .xz if possible and also a hash of some kind for verification regardless of archive type

benbrandt:
Since we've already decided to start implementation on this, I am going to go ahead and marge the current version as a draft 😄 
We can do further modifications to make it better match the final specification as we do testing before it moves to Preview stage

docs/rfds/acp-agent-registry.mdx
nikomatsakis: A proxy is shipped exactly the same was an agent. It's a self-contained binary -- but of course it could be an npx or pipx package.

ignatov: Great

PR #297, opened 2025-12-04
docs: Better document _meta field type and usage

codefromthecrypt:
Update extensibility documentation with:
- Type annotation for _meta field: { [key: string]: unknown }
- W3C trace context propagation conventions (traceparent, tracestate, baggage)
- Example showing traceparent in _meta

Follow-up to https://github.com/agentclientprotocol/agent-client-protocol/pull/276

### DIFF
--- a/docs/protocol/extensibility.mdx
+++ b/docs/protocol/extensibility.mdx
@@ -7,7 +7,7 @@ The Agent Client Protocol provides built-in extension mechanisms that allow impl
 
 ## The `_meta` Field
 
-All types in the protocol include a `_meta` field that implementations can use to attach custom information. This includes requests, responses, notifications, and even nested types like content blocks, tool calls, plan entries, and capability objects.
+All types in the protocol include a `_meta` field with type `{ [key: string]: unknown }` that implementations can use to attach custom information. This includes requests, responses, notifications, and even nested types like content blocks, tool calls, plan entries, and capability objects.
 
 ```json
 {
@@ -23,11 +23,18 @@ All types in the protocol include a `_meta` field that implementations can use t
       }
     ],
     "_meta": {
+      "traceparent": "00-80e1afed08e019fc1110464cfa66635c-7a085853722dc6d2-01",
       "zed.dev/debugMode": true
     }
   }
 }
 ```
+
+Clients may propagate fields to the agent for correlation purposes, such as `requestId`. The following root-level keys in `_meta` **SHOULD** be reserved for [W3C trace context](https://www.w3.org/TR/trace-context/) to guarantee interop with existing MCP implementations and OpenTelemetry tooling:
+
+- `traceparent`
+- `tracestate`
+- `baggage`
 
 Implementations **MUST NOT** add any custom fields at the root of a type that's part of the specification. All possible names are reserved for future protocol versions.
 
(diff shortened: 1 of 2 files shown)


docs/protocol/extensibility.mdx
```diff
@@ -23,12 +23,19 @@ All types in the protocol include a `_meta` field that implementations can use t
       }
     ],
     "_meta": {
+      "traceparent": "00-80e1afed08e019fc1110464cfa66635c-7a085853722dc6d2-01",
```
codefromthecrypt, line 26: this wasn't in the RFD, but figured it can re-enforce below. happy to remove it though

docs/protocol/extensibility.mdx
```diff
@@ -7,7 +7,7 @@ The Agent Client Protocol provides built-in extension mechanisms that allow impl
 
 ## The `_meta` Field
 
-All types in the protocol include a `_meta` field that implementations can use to attach custom information. This includes requests, responses, notifications, and even nested types like content blocks, tool calls, plan entries, and capability objects.
+All types in the protocol include a `_meta` field with type `{ [key: string]: unknown }` that implementations can use to attach custom information. This includes requests, responses, notifications, and even nested types like content blocks, tool calls, plan entries, and capability objects.
```
cliffhall, line 10: This elaboration is TypeScript-centric. I think we can make it more JSON-centric.

```suggestion
All types in the protocol include a `_meta` field that implementations can use to attach custom information. It is an object whose keys are strings and values can be any other JSON value. This includes requests, responses, notifications, and even nested types like content blocks, tool calls, plan entries, and capability objects.
```

docs/protocol/extensibility.mdx
cliffhall: The existing example of `zed.dev/debugMode` illustrates an important part of using `_meta`, which is making the key unique to your server to avoid naming conflicts.

Even with the follow-on explanation, I fear the presence of the `traceparent` key in the example could be confusing, especially if someone only checks the example and doesn't read on to find out that this is the execption to the rule — that we're advocating a carve-out for an existing tech for compatibility.

docs/protocol/extensibility.mdx
```diff
@@ -23,12 +23,19 @@ All types in the protocol include a `_meta` field that implementations can use t
...
     ],
     "_meta": {
+      "traceparent": "00-80e1afed08e019fc1110464cfa66635c-7a085853722dc6d2-01",
       "zed.dev/debugMode": true
     }
   }
 }
 ```
 
+Clients may propagate fields to the agent for correlation purposes, such as `requestId`. The following root-level keys in `_meta` **SHOULD** be reserved for [W3C trace context](https://www.w3.org/TR/trace-context/) to guarantee interop with existing MCP implementations and OpenTelemetry tooling:
```
cliffhall, line 33:
```suggestion
A best practice for naming keys in the `_meta` object is to uniquify it with your domain or app name, e.g,, `"zed.dev/debugMode": true` in the example above. This helps to avoid naming conflicts. 

In some cases, however, for compatibility with existing and related technologies, it may not be possible to uniquify the key. In those cases, a request to reserve their keys will be included in this spec.

### OpenTelemetry key reservations
The following root-level keys in `_meta` **SHOULD** be reserved for [W3C trace context](https://www.w3.org/TR/trace-context/) to guarantee interop with existing MCP implementations and OpenTelemetry tooling:

```

cliffhall:
LOL. Somehow because it showed up in my notifications, I thought this was an MCP Spec PR. Feel free to ignore the above if it makes no sense here.

PR #298, opened 2025-12-04
docs(rfd): Draft: Agent Telemetry Export

codefromthecrypt:
Define how agents export telemetry (logs, metrics, traces) to clients without tunneling over the ACP transport.

 - Clients run a local OTLP receiver and inject `OTEL_EXPORTER_*` env vars when launching agents
 - Keeps telemetry out-of-band to avoid head-of-line blocking on stdio
 - Agents use standard OpenTelemetry SDKs with auto-configuration
 - No credential sharing: client proxies telemetry to backends with its own credentials
 - Complements meta-propagation RFD (trace context correlation)

Implementation proposes a new `docs/protocol/observability.mdx` covering both client and agent author guidance.

extracted from https://github.com/agentclientprotocol/agent-client-protocol/pull/295#issuecomment-3609493050

### DIFF
--- a/docs/rfds/agent-telemetry-export.mdx
+++ b/docs/rfds/agent-telemetry-export.mdx
@@ -3,6 +3,7 @@ title: "Agent Telemetry Export"
 ---
 
 - Author(s): [@codefromthecrypt](https://github.com/codefromthecrypt)
+- Champion: [@benbrandt](https://github.com/benbrandt)
 
 ## Elevator pitch
 


codefromthecrypt:
Thanks I will report back with an impl!

morgante:
@codefromthecrypt Did you end up implementing this anywhere?

codefromthecrypt:
@morgante TL'DR' not yet. there's been a lot of related housekeeping work in goose in both otel and ACP in prep of this, but not all of it is done.

PR #303, opened 2025-12-05
feat(unstable): Draft implementation of `$/cancel_request` notification

benbrandt:
Adds basic schema primitives and documentation for the cancel request notification.

### DIFF
--- a/src/bin/generate.rs
+++ b/src/bin/generate.rs
@@ -875,6 +875,7 @@ starting with '$/' it is free to ignore the notification."
             }
         }
 
+        #[cfg(feature = "unstable_cancel_request")]
         fn protocol_method_doc(&self, method_name: &str) -> &String {
             match method_name {
                 "$/cancel_request" => self.protocol.get("CancelRequestNotification").unwrap(),


docs/rfds/request-cancellation.mdx
```diff
@@ -20,7 +20,7 @@ This creates the following inconveniences:
 
 ## What we propose to do about it
 
-Implement an **optional** `$/cancelRequest` notification method (inspired by the Language Server Protocol) that uses JSON-RPC 2.0 notification format, allowing either party (client or agent) to cancel any outstanding request by its ID.
+Implement an **optional** `request/cancel` notification method (inspired by the Language Server Protocol) to both Agent and Client that uses JSON-RPC 2.0 notification format, allowing either party (client or agent) to cancel any outstanding request by its ID.
```
benbrandt: @nerzhulart I changed this to `request/cancel` because it seemed to fit better with the existing methods names.

If we want the `$/` prefix, we can, but then it would still be `$/cancel_request`

The reason to do that would be if we want the same blanket LSP caveat that $ are not implementable by all sides: https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#dollarRequests

We kind of have this already by having it be a capability. If we go the LSP route, we wouldn't need a capability, however neither side would know if their notifications are supported or not. But since you can already send one and the server could have responded before seeing the notification, maybe this is fine.

So basically the question is: how significant is it to the caller side to know if their cancellation requests will be supported or not?

docs/protocol/draft/schema.mdx
```diff
@@ -158,6 +158,54 @@ The client should disconnect, if it doesn't support this version.
...
+<a id="request-cancel"></a>
+### <span class="font-mono">request/cancel</span>
+
+**UNSTABLE**
+
+This capability is not part of the spec yet, and may be removed or changed at any point.
+
+Cancels an ongoing request.
+
+This is a notification sent by the agent to cancel any ongoing request.
```
nerzhulart: I suspect `agent` and `client` are inverted in this block

nerzhulart: BTW don't you want to make this cancellation block common for both client and agent? At least it will be easier to make changes in future

docs/rfds/request-cancellation.mdx
nerzhulart: I don't mind to change the name at all. The idea was to have some kind of a marker that this method is a meta-method on the JsonPRC level rather than a part the ACP domain model

nerzhulart: But actually the name doesn't affect the implementation significantly, so if you think that the new name is better I'll change it as well in the Kotlin SDK once the PR is merged

src/bin/generate.rs
```diff
@@ -821,6 +822,10 @@ and control access to resources."
                 "session/prompt" => self.agent_methods.get("PromptRequest").unwrap(),
                 "session/cancel" => self.agent_methods.get("CancelNotification").unwrap(),
                 "session/set_model" => self.agent_methods.get("SetSessionModelRequest").unwrap(),
+                "request/cancel" => self
+                    .client_methods
```
nerzhulart: shouldn't be `agent_methods` since the methods is for the agent?

benbrandt: Good catch. Explains the docs problem earlier

docs/protocol/draft/schema.mdx
benbrandt: It is the same type under the hood.

The reason why it is on both is because of how the docs are currently structured, it shows up nicely with all of the other methods.

But I could have a "generic" methods section I guess.. I'll see what i can do here, but I guess my thought was these are still "methods" that the agent and client need to implement, which is why I appended them to the current lists.

I felt duplication of one comment (the rest is all shared) was a small price for the implementer to see in one place all of the methods expected of them as an agent or client

docs/rfds/request-cancellation.mdx
benbrandt: I kind of go back and forth. The main benefit i see of doing `$` is having it as a clear namespace for notifications that may be ignored because they may not always be implementable (and therefore may also remove the need for a capability)

PR #316, opened 2025-12-07
docs(rfd): Add session usage and context status RFD

ahmedhesham6:
Proposes standardized tracking of token usage, cost estimation, and context window status across ACP implementations.

- Token usage reported in PromptResponse (per-turn data)
- Context window and cost reported in session/status (session state)

### DIFF
--- a/docs/rfds/session-usage.mdx
+++ b/docs/rfds/session-usage.mdx
@@ -3,6 +3,7 @@ title: "Session Usage and Context Status"
 ---
 
 - Author(s): [@ahmedhesham6](https://github.com/ahmedhesham6)
+- Champion: [@benbrandt](https://github.com/benbrandt)
 
 ## Elevator pitch
 
@@ -47,16 +48,19 @@ We propose separating usage tracking into two distinct concerns:
 2. **Context window and cost** - Reported via `session/update` notifications with `sessionUpdate: "usage_update"` (session state)
 
 This separation reflects how users consume this information:
+
 - Token counts are tied to specific turns and useful immediately after a prompt
 - Context window and cost are cumulative session state that agents push proactively when available
 
 Agents send context updates at appropriate times:
+
 - On `session/new` response (if agent can query usage immediately)
 - On `session/load` / `session/resume` (for resumed/forked sessions)
 - After each `session/prompt` response (when usage data becomes available)
 - Anytime context window state changes significantly
 
 This approach provides flexibility for different agent implementations:
+
 - Agents that support getting current usage without a prompt can immediately send updates when creating, resuming, or forking chats
 - Agents that only provide usage when actively prompting can send updates after sending a new prompt
 
@@ -211,6 +215,7 @@ Different users care about different things at different times:
 - **Cumulative cost**: Session-level state that agents push when available
 
 Separating them allows:
+
 - Cleaner data model where per-turn data stays in turn responses
 - Agents to push context updates proactively when data becomes available
 - Clients to receive updates reactively without needing to poll
@@ -218,6 +223,7 @@ Separating them allows:
 ### Why is cost in session/update instead of PromptResponse?
 
 Cost is cumulative session state, similar to context window:
+
 - Users want to track total spending, not just per-turn costs
 - Keeps `PromptResponse` focused on per-turn token breakdown
 - Both cost and context window are session-level metrics that belong together
@@ -233,14 +239,15 @@ The context update notification provides everything needed:
 
 **Recommended client behavior:**
 
-| Percentage | Action |
-|------------|--------|
-| < 75% | Normal operation |
-| 75-90% | Yellow indicator, suggest "Context filling up" |
-| 90-95% | Orange indicator, recommend "Start new session or summarize" |
-| > 95% | Red indicator, warn "Next prompt may fail - handoff recommended" |
+| Percentage | Action                                                           |
+| ---------- | ---------------------------------------------------------------- |
+| < 75%      | Normal operation                                                 |
+| 75-90%     | Yellow indicator, suggest "Context filling up"                   |
+| 90-95%     | Orange indicator, recommend "Start new session or summarize"     |
+| > 95%      | Red indicator, warn "Next prompt may fail - handoff recommended" |
 
 Clients can also:
+
 - Offer "Compact context" or "Summarize conversation" actions
 - Auto-suggest starting a new session
 - Implement automatic handoff when approaching limits
--- a/docs/docs.json
+++ b/docs/docs.json
@@ -112,7 +112,8 @@
               "rfds/meta-propagation",
               "rfds/session-info-update",
               "rfds/agent-telemetry-export",
-              "rfds/proxy-chains"
+              "rfds/proxy-chains",
+              "rfds/session-usage"
             ]
           },
           { "group": "Preview", "pages": [] },
--- a/docs/updates.mdx
+++ b/docs/updates.mdx
@@ -4,6 +4,20 @@ description: Updates and announcements about the Agent Client Protocol
 rss: true
 ---
 
+<Update label="January 1, 2026" tags={["RFD"]}>
+## Session Usage RFD moves to Draft stage
+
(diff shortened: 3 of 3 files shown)


docs/rfds/session-usage-context-status.mdx
```diff
@@ -0,0 +1,288 @@
...
+  "jsonrpc": "2.0",
+  "id": 2,
+  "result": {
+    "sessionId": "sess_abc123",
+    "status": "idle",
+    "context_window": {
+      "size": 200000,
+      "used": 53000,
+      "percentage": 26.5,
+      "remaining": 147000
```
versecafe: are these two fields needed, they are fully derivable from `size` and `used`

docs/rfds/session-usage-context-status.mdx
```diff
@@ -0,0 +1,288 @@
...
+  }
+}
+```
+
+#### Context Window Fields (optional)
+
+- `size` (number, required) - Total context window size in tokens
+- `used` (number, required) - Tokens currently in context
+- `percentage` (number, required) - Percentage used (0-100)
+- `remaining` (number, required) - Tokens remaining
```
versecafe: ditto on the derived fields from `size` and `used` are these fields needed

docs/rfds/session-usage.mdx
```diff
@@ -0,0 +1,288 @@
...
+- `remaining` (number, required) - Tokens remaining
+
+#### Cost Fields (optional)
+
+- `amount` (number, required) - Total cumulative cost for session
+- `currency` (string, required) - ISO 4217 currency code (e.g., "USD", "EUR")
+
+### Design Principles
+
+1. **Separation of concerns** - Token usage is per-turn data, context window and cost are session state
```
versecafe, line 154: how will this work with subagent "tools" that aren't performing full turns but are actively updating token usage frequently

docs/rfds/session-usage-context-status.mdx
ahmedhesham6: They’re redundant mathematically, but they carry the agent’s own calculation/rounding and simplify client work

benbrandt:
@josevalim @SteffenDE I'd love to get your input on this one since you were looking into this a bit

SteffenDE:
We like the idea and we definitely want to have a way to do this in ACP! For us, the most important part is the current usage (in percent). Including a `usage` in prompt responses feels like a no-brainer, but since an ACP prompt often consists of multiple agent turns, wrappers like Claude-Code-ACP would need to accumulate the different tokens from the turns. I think that's alright though. As mentioned in https://github.com/agentclientprotocol/agent-client-protocol/pull/316#discussion_r2600186578, one needs to be careful about usage data from subagents, as those should not be included, or optionally provided separately.

The PR proposes a new `session/status` method. I'm not sure if agents like Claude Code have proper APIs to query the current status at any time, so a different idea would be to only send the current usage information (current tokens, max tokens, percent) as part of `session/update` notifications only. An agent that supports getting the current usage without a prompt may then immediately send the update when creating a new chat, resuming a chat, forking a chat, etc., similar to how the available command updates are sent. An agent that only provides usage when actively prompting could only start sending the updates after sending a new prompt. That might mean that when resuming a chat, the client UI cannot immediately show the usage, but it allows more flexibility for agents.

ahmedhesham6:
@SteffenDE Thanks a lot for the detailed explanation and context 🙏
Just to confirm I understood you correctly — is this roughly what you’re suggesting for the session/update approach?

```
{
  "jsonrpc": "2.0",
  "method": "session/update",
  "params": {
    "sessionId": "sess_abc123",
    "update": {
      "sessionUpdate": "context",
      "used": 53000,
      "size": 200000,
      "percentage": 26.5
    }
  }
}
```
If this matches what you had in mind, I can adjust the RFD in that direction.

SteffenDE:
@ahmedhesham6 yes! I'd wait before changing things though, since I'm not a maintainer here and basically just stating my opinion :D

ahmedhesham6:
What do you think @benbrandt?

benbrandt:
Yeah I think something simple to start would be great. As @SteffenDE mentioned, support for this will likely vary wildly (and we've also seen mixed support of even these basic metrics within the same agent lol) 

I think we should let this be driven by the agent, as they will likely get the information from the provider and may forward it, but might not hold on to it. Requiring them to have the data at all points might be too much... So I'd opt for a simple way to report the basic information we feel we need, and go from there

docs/rfds/session-usage-context-status.mdx
```diff
@@ -0,0 +1,326 @@
...
+  }
+}
+```
+
+#### Usage Fields
+
+- `total_tokens` (number, required) - Sum of all token types across session
+- `input_tokens` (number, required) - Total input tokens across all turns
+- `output_tokens` (number, required) - Total output tokens across all turns
+- `reasoning_tokens` (number, optional) - Total reasoning tokens (for o1/o3 models)
```
benbrandt: since in ACP we usually refer to this as `thought` I wonder if we could align that?

docs/rfds/session-usage-context-status.mdx
```diff
@@ -0,0 +1,326 @@
...
+    }
+  }
+}
+```
+
+#### Context Window Fields (required)
+
+- `used` (number, required) - Tokens currently in context
+- `size` (number, required) - Total context window size in tokens
+- `percentage` (number, required) - Percentage used (0-100)
```
benbrandt: Do we send this just to save a calculation on the client? since we allow them to calculate remaining, maybe we let them do this too?

docs/rfds/session-usage.mdx
```diff
@@ -0,0 +1,326 @@
...
+
+- `used` (number, required) - Tokens currently in context
+- `size` (number, required) - Total context window size in tokens
+- `percentage` (number, required) - Percentage used (0-100)
+
+Note: Clients can compute `remaining` as `size - used` if needed.
+
+#### Cost Fields (optional)
+
+- `cost` (object, optional) - Cumulative session cost
```
benbrandt, line 127: naming nitpick: it seems weird that this is part of a "context" update. 
I wonder if all of this is just `usage` from a conceptual point of view?
And roughly the same data can be sent at the end of the turn, with mid-turn updates? So kind of merge these?

It seems you want to distinguish between turn usage vs total usage. Which makes sense, but I wonder if we can distinguish then between turn vs session usage?

docs/rfds/session-usage-context-status.mdx
```diff
@@ -0,0 +1,326 @@
...
+- Implement automatic handoff when approaching limits
+
+### Why does the agent calculate percentage instead of the client?
+
+Agent knows its model best:
+
+- Agent knows exact context window size (varies by model)
+- Agent knows how it counts tokens (different tokenizers)
+- Agent knows about special tokens, system messages, etc.
+- Client can still recalculate if needed (all raw data provided)
```
benbrandt: I dont' really understand these arguments... Unless percentage will somehow be different than `used / total` (which would be weird) I feel the agent gets all of this by controlling the token counts already

docs/rfds/session-usage.mdx
ahmedhesham6: @benbrandt Great point on the naming! You're right that "context" is a bit awkward here.

I think we're aligned on the core idea - distinguishing between **turn usage** vs **session usage**:

- **Turn usage** (in `PromptResponse`): Token breakdown for that specific turn - input, output, thought, cached tokens. This is the "what just happened" data.

- **Session usage** (via `session/update`): Cumulative state - total tokens consumed, context window utilization, cost. This is the "where are we now" data that agents push when available.

The `session/update` notification could use `sessionUpdate: "usage_update"` instead of `"context_update"` to better reflect this. The payload would still include context window info (`used`, `size`) alongside cumulative cost, but framing it as "usage" makes more conceptual sense.

Would renaming to `sessionUpdate: "usage_update"` address the naming concern? Happy to update the RFD if that direction works.

benbrandt: Yeah I think we're on the same page here 👍🏻 thanks!

benbrandt (approved):
This looks great, thanks for the iterations here! ❤️

PR #277, opened 2025-11-28
add session/fork

SteffenDE:
Attempt to implement the session/fork RFD (https://github.com/agentclientprotocol/agent-client-protocol/pull/233).

Disclaimer: I gave the RFD to Claude and let it implement the code. I reviewed everything and it looks reasonable to me, but I'm not a Rust expert :)

I did encounter one issue when I tried to use the generated schema in the typescript-sdk, but I don't think it is related: in the schema.ts, the response types are missing, e.g. `NewSessionResponse`. In fact, I get the same error trying to use 0.7.0, so it seems to be a general incompatibility.

PR #324, opened 2025-12-09
feat(unstable): Draft implementation of session/resume

SteffenDE:
Similar to https://github.com/agentclientprotocol/agent-client-protocol/pull/277 / https://github.com/agentclientprotocol/agent-client-protocol/pull/311.

I did not regenerate the schema files for now.

### DIFF
--- a/src/agent.rs
+++ b/src/agent.rs
@@ -2469,6 +2469,7 @@ impl ClientRequest {
 #[serde(untagged)]
 #[schemars(inline)]
 #[non_exhaustive]
+#[expect(clippy::large_enum_variant)]
 pub enum AgentResponse {
     InitializeResponse(InitializeResponse),
     AuthenticateResponse(#[serde(default)] AuthenticateResponse),


src/agent.rs
```diff
@@ -732,6 +732,142 @@ impl ForkSessionResponse {
...
+///
+/// This capability is not part of the spec yet, and may be removed or changed at any point.
+///
+/// Response from resuming an existing session.
+#[cfg(feature = "unstable_session_resume")]
+#[derive(Default, Debug, Clone, Serialize, Deserialize, JsonSchema, PartialEq, Eq)]
+#[schemars(extend("x-side" = "agent", "x-method" = SESSION_RESUME_METHOD_NAME))]
+#[serde(rename_all = "camelCase")]
+#[non_exhaustive]
+pub struct ResumeSessionResponse {
```
SteffenDE, line 809: We should probably discuss this separately, but I think it might be useful if agents are allowed to return a different session ID for resume if the upstream agent works that way.

SteffenDE: After thinking about this: it makes more sense to use fork in that case. If the agent cannot continue a session without requiring a new id, it should just not implement resume!

benbrandt:
@SteffenDE for some reason I can't push to your PRs.... I'm guessing the checkbox for "allow maintainers to edit" isn't checked?

benbrandt:
Solved it by doing it from your repo

SteffenDE:
@benbrandt it's weird because I don't have that checkbox on this PR, but I remember seeing it in the past in different repositories...

PR #326, opened 2025-12-10
docs(rfd): fork options

SteffenDE:
We propose to change the `session/fork` method to accept the same options as `session/load`. This eases the implementation in agents that do not persistently store all the configuration when a session is not active any more.

### DIFF
--- a/docs/rfds/session-fork.mdx
+++ b/docs/rfds/session-fork.mdx
@@ -61,7 +61,7 @@ Then the client would be able to request a fork of the given session:
 }
 ```
 
-The request accepts the same same options as `session/load`, such as `cwd` and `mcpServers`.
+The request expects the same options as `session/load`, such as `cwd` and `mcpServers`.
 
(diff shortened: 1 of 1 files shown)


SteffenDE:
cc @josevalim

docs/rfds/session-fork.mdx
```diff
@@ -46,22 +46,27 @@ We propose to add a new "session/fork" method. Agents must declare this option i
...
-    "sessionId": "sess_789xyz"
+    "sessionId": "sess_789xyz",
+    "cwd": "...",
+    "mcpServers": [...]
   }
 }
 ```
 
-And the agent would respond with optional data such as config options, the same as `session/load`.
+The request accepts the same same options as `session/load`, such as `cwd` and `mcpServers`.
```
josevalim:
```suggestion
The request expects the same options as `session/load`, such as `cwd` and `mcpServers`.
```

PR #340, opened 2025-12-15
RFD for agent-guided user selection feature

akhil-vempali:
Proposes `session/select` request allowing agents to present interactive menus with prompts and markdown-renderable options. Supports single/multi-select modes with optional free-text input for gathering structured user input and exposing agent capabilities in a discoverable way.

### DIFF
--- a/docs/rfds/agent-guided-user-selection.mdx
+++ b/docs/rfds/agent-guided-user-selection.mdx
@@ -1,0 +1,259 @@
+---
+title: "Agent-Guided User Selection"
+---
+
+Author(s): [akhil-vempali](https://github.com/akhil-vempali)
+
+## Elevator pitch
+
+> What are you proposing to change?
+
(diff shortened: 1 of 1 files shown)


benbrandt:
@akhil-vempali thanks for this. We are discussing in #330 the idea of implementing MCP elicitations which would likely help with case?

The idea is that most clients will need to use elicitations anyway from MCP servers, so we could leverage the same mechanism?

akhil-vempali:
Hey @benbrandt yep I think that should do the work as well. Is there a RFD specifically for elicitation? The one you shared is only for auth?

benbrandt:
@akhil-vempali you're right it got removed for now. EIther way, I think we'd rather have an MCP elicitation story in general, and maybe offer the same interface outside of tool calls. Mind adjusting the RFD to go in that direction?

benbrandt:
closing in favor of #376

PR #359, opened 2025-12-31
RFD: Bring Symposium into ACP organization

nikomatsakis:
## Elevator pitch

Bring the [Symposium project](https://github.com/symposium-dev/symposium) into the ACP organization. Built on ACP's proxy architecture, Symposium enables library ecosystems where packages come with their own agent integration, creating extensible agents that users can customize with domain-specific capabilities. Symposium currently focuses on Rust, but the vision applies to any language ecosystem. It will expand to other languages over time.

Symposium currently consists of two things:

1. An ACP proxy for Rust development. It orchestrates specialized sub-proxies and is expected to evolve to intelligently select them based on current dependencies, user chat contents, and other factors. The proxy can be used from any ACP-supported editor.
2. A VSCode extension (available on the [VSCode](https://marketplace.visualstudio.com/items?itemName=symposium-dev.symposium) and [OpenVSIX](https://open-vsx.org/extension/symposium-dev/symposium) marketplaces) that provides a generic ACP front-end, allowing users to easily experiment with Symposium using any ACP agent found on the [ACP registry](https://github.com/agentclientprotocol/registry).

### DIFF
--- a/docs/docs.json
+++ b/docs/docs.json
@@ -104,6 +104,7 @@
           {
             "group": "Draft",
             "pages": [
+              "rfds/symposium",
               "rfds/session-list",
               "rfds/session-config-options",
               "rfds/session-fork",
--- a/docs/rfds/symposium.mdx
+++ b/docs/rfds/symposium.mdx
@@ -1,0 +1,134 @@
+---
+title: "Symposium: Self-Integrating Library Ecosystems"
+---
+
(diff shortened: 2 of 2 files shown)


nikomatsakis:
I've decided not to do this for the time being. Instead, we'll upstream ideas as they prove their worth.

PR #364, opened 2026-01-08
split proxy + MCP-over-ACP RFDs

nikomatsakis:
This splits out the proxy support from MCP-over-ACP, as they are conceptually distinct, and produces two independent RFDs.

### DIFF
--- a/docs/rfds/mcp-over-acp.mdx
+++ b/docs/rfds/mcp-over-acp.mdx
@@ -70,16 +70,16 @@ sequenceDiagram
     Agent-->>Client: session created
 
     Client->>Agent: prompt ("analyze this codebase")
-    
+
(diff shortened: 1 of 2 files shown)


codefromthecrypt (commented):
this is interesting work. I will share it with the transports interest group in MCP since this is about adding a new transport and has overlap with other routine things like the wasm question.

one nit is maybe we should rename the proxy-chain doc as well?

(this isn't a full review, just a note as I looked at this briefly)

benbrandt (approved):

nikomatsakis:
> maybe we should rename the proxy-chain doc as well?

@codefromthecrypt I debated about that. I wanted to keep the URL the same, but I am trying to rebrand to "agent extensions", so it might make sense.

PR #376, opened 2026-01-12
RFD: Add Elicitation specification for structured user input

yordis:
Signed-off-by: Yordis Prieto <yordis.prieto@gmail.com>

### DIFF
--- a/docs/rfds/elicitation.mdx
+++ b/docs/rfds/elicitation.mdx
@@ -2,7 +2,8 @@
 title: "Elicitation: Structured User Input During Sessions"
 ---
 
-Author(s): [@yordis](https://github.com/yordis)
+- Author(s): [@yordis](https://github.com/yordis)
+- Champion: [@benbrandt](https://github.com/benbrandt)
 
 ## Elevator pitch
 
@@ -72,6 +73,7 @@ Clients can:
 This proposal follows MCP's draft elicitation specification. See [MCP Elicitation Specification](https://modelcontextprotocol.io/specification/draft/client/elicitation) for detailed guidance. ACP uses the same JSON Schema constraint approach and capability model, adapted for our session/turn-based architecture.
 
 Key differences from MCP:
+
 - MCP elicitation is tool-call-scoped; ACP elicitation is session-scoped
 - ACP uses `session/elicitation` method; MCP uses `elicitation/create`
 - ACP must integrate with existing Session Config Options (which also use schema constraints)
@@ -94,9 +96,15 @@ Agents send elicitation requests when they need information from the user. This 
         "title": "Refactoring Strategy",
         "description": "Choose how aggressively to refactor",
         "oneOf": [
-          { "const": "conservative", "title": "Conservative - Minimal changes" },
+          {
+            "const": "conservative",
+            "title": "Conservative - Minimal changes"
+          },
           { "const": "balanced", "title": "Balanced (Recommended)" },
-          { "const": "aggressive", "title": "Aggressive - Maximum optimization" }
+          {
+            "const": "aggressive",
+            "title": "Aggressive - Maximum optimization"
+          }
         ],
         "default": "balanced"
       }
@@ -183,6 +191,7 @@ Following MCP's approach (specifically [SEP-1036](https://modelcontextprotocol.i
 This distinction is reflected in the client capabilities model, allowing clients to declare support for one or both modalities.
 
 **Normative requirements:**
+
 - Clients declaring the `elicitation` capability MUST support at least one mode (`form` or `url`).
 - Agents MUST NOT send elicitation requests with modes that are not supported by the client.
 - For URL mode, the `url` parameter MUST contain a valid URL.
@@ -195,6 +204,7 @@ Aligning with [MCP's draft elicitation specification](https://modelcontextprotoc
 **Supported primitive types:**
 
 1. **String Schema**
+
 ```json
 {
   "type": "string",
@@ -207,9 +217,11 @@ Aligning with [MCP's draft elicitation specification](https://modelcontextprotoc
   "default": "user@example.com"
 }
 ```
+
 Supported formats: `email`, `uri`, `date`, `date-time`
 
 2. **Number Schema**
+
 ```json
 {
   "type": "number",
@@ -220,9 +232,11 @@ Supported formats: `email`, `uri`, `date`, `date-time`
   "default": 50
 }
 ```
+
 Also supports `"type": "integer"` for whole numbers.
 
 3. **Boolean Schema**
+
 ```json
 {
   "type": "boolean",
@@ -235,6 +249,7 @@ Also supports `"type": "integer"` for whole numbers.
 4. **Enum Schema** (for selections)
 
 Single-select enum (without titles):
+
 ```json
 {
   "type": "string",
@@ -246,6 +261,7 @@ Single-select enum (without titles):
 ```
 
 Single-select enum (with titles):
+
 ```json
 {
   "type": "string",
@@ -261,6 +277,7 @@ Single-select enum (with titles):
 ```
 
 Multi-select enum (without titles):
+
 ```json
 {
   "type": "array",
@@ -277,6 +294,7 @@ Multi-select enum (without titles):
 ```
 
 Multi-select enum (with titles):
+
 ```json
 {
   "type": "array",
@@ -296,6 +314,7 @@ Multi-select enum (with titles):
 ```
 
 **Request schema structure:**
+
 ```json
 "requestedSchema": {
   "type": "object",
@@ -316,6 +335,7 @@ Multi-select enum (with titles):
 ```
 
 **Not supported** (to simplify client implementation):
+
 - Complex nested objects/arrays (beyond enum arrays)
 - Conditional validation
 - Custom formats beyond the supported list
@@ -329,6 +349,7 @@ Clients use this schema to generate appropriate input forms, validate user input
 The agent sends a `session/elicitation` request when it needs information from the user:
 
 **Form mode example:**
+
 ```json
 {
   "jsonrpc": "2.0",
@@ -359,6 +380,7 @@ The agent sends a `session/elicitation` request when it needs information from t
 ```
 
 **URL mode example:**
+
 ```json
 {
   "jsonrpc": "2.0",
@@ -381,6 +403,7 @@ The client presents the elicitation UI to the user. For form mode, the client ge
 Elicitation responses use a three-action model (following MCP) to clearly distinguish between different user actions:
 
 **Accept** - User explicitly approved and submitted with data:
+
 ```json
 {
   "jsonrpc": "2.0",
@@ -395,6 +418,7 @@ Elicitation responses use a three-action model (following MCP) to clearly distin
 ```
 
 **Decline** - User explicitly declined the request:
+
 ```json
 {
   "jsonrpc": "2.0",
@@ -406,6 +430,7 @@ Elicitation responses use a three-action model (following MCP) to clearly distin
 ```
 
 **Cancel** - User dismissed without making an explicit choice (closed dialog, pressed Escape, etc.):
+
 ```json
 {
   "jsonrpc": "2.0",
@@ -419,6 +444,7 @@ Elicitation responses use a three-action model (following MCP) to clearly distin
 For URL mode elicitation, the response with `action: "accept"` indicates that the user consented to the interaction. It does not mean the interaction is complete—the interaction occurs out-of-band and the client is not aware of the outcome until the agent sends a completion notification.
 
 Agents should handle each state appropriately:
+
 - **Accept**: Process the submitted data
 - **Decline**: Handle explicit decline (e.g., use default, offer alternatives)
 - **Cancel**: Handle dismissal (e.g., use default, prompt again later)
@@ -513,10 +539,12 @@ Following MCP, agents MAY send a `notifications/elicitation/complete` notificati
 ```
 
 Agents sending notifications:
+
 - MUST only send the notification to the client that initiated the elicitation request
 - MUST include the `elicitationId` established in the original request
 
 Clients:
+
 - MUST ignore notifications referencing unknown or already-completed IDs
 - MAY use this notification to automatically retry requests, update UI, or continue an interaction
 - SHOULD provide manual controls for the user to retry or cancel if the notification never arrives
@@ -588,10 +616,12 @@ Clients declare elicitation support during the `initialize` phase via `ClientCap
 ```
 
 **Capability structure:**
+
 - `elicitation.form` - Present if the client can render form UI from restricted JSON Schema (strings, numbers, integers, booleans, enums)
 - `elicitation.url` - Present if the client can open URLs for out-of-band flows (OAuth, payments, credential collection)
 
 **Example: Headless client (no browser access):**
+
 ```json
 "elicitation": {
   "form": {}
@@ -599,6 +629,7 @@ Clients declare elicitation support during the `initialize` phase via `ClientCap
 ```
 
 **Example: Simple terminal with URL support only:**
+
 ```json
 "elicitation": {
   "url": {}
@@ -606,6 +637,7 @@ Clients declare elicitation support during the `initialize` phase via `ClientCap
 ```
 
 **Example: Full-featured client:**
+
 ```json
 "elicitation": {
   "form": {},
@@ -614,6 +646,7 @@ Clients declare elicitation support during the `initialize` phase via `ClientCap
 ```
 
 This structure:
+
 1. Allows clients to declare partial support based on their environment
 2. Enables agents to pass capabilities along to MCP servers they connect to
 3. Maps cleanly to MCP's elicitation capability model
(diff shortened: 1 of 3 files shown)


ignatov:
Hey! Thanks for contributing, is it true, that we can handle things like that https://github.com/orgs/agentclientprotocol/discussions/371 with this feature?

phil65:
Just for some context, OpenCodes question tool schema:
https://github.com/anomalyco/opencode/pull/7268/changes#diff-2036e0c76252554ccff3dfc803b49917c90430275cadc8f9962ba528780c4e79
Claude codes question tool schema:
https://platform.claude.com/docs/en/agent-sdk/python#ask-user-question
I guess these question tools as well as MCP elicitaiton are the most common use cases for this feature (+ perhaps some onboarding flows).
This is one of the biggest gaps in ACP currently I think, so +1 from me for this RFD :)

yordis:
Hey @ignatov these type of spec should allow for that type of "question" form to happen. In fact, my existing ACP needs is all about such feature.

I am trying to write the spec AND figure out if I can make Zed GUI to implement it.

Overall, the intent is to allow structure input from the user such as Questionnaire, or buttons for specific auctions (at least that is my immediate need).

docs/rfds/elicitation.mdx
```diff
@@ -0,0 +1,295 @@
...
+
+This constraint list can expand in future versions based on community feedback.
+
+### User Response
+
+When a user responds to an elicitation request, the response is included in the next turn request:
+
+```json
+{
+  "method": "session/turn",
```
benbrandt: Is this a new method? Or what are you referring to here? I think it would make sense to potentially model this after how we do permission requests (there is an eliciation method with request/response pattern)

and then I think we need someway to associate this with content in the thread (maybe the client just puts it in there once it receives such a request)

yordis:
@benbrandt is it OK to ask to ignore the Rust changes for now, I know it could be annoying, but I am trying to make Zed to work to see the full picture, I just realized that I committed those changes as well while trying to fix the request/response situation.

Since I am in such active development, and I want to see Zed working, the burden is on you to ignore those files and only focus on the markdown until we are ready to merge.

Otherwise, totally cool, I create another branch for myself, just make it a bit more difficult since it requires to switch between them and synchronize a bit more.

ignatov:
@benbrandt I think that we have to work on that in the next wave

yordis:
@ignatov tomorrow I am back home so I can continue the work, I got distracted by trying to actually make a Zed GUI that allow us to do OpenCode-style of questionnaire as a PoC.

Let me know whatever you would like to see happening, I am available in 24hrs

docs/rfds/elicitation.mdx
```diff
@@ -0,0 +1,403 @@
...
+The mechanism would:
+
+1. **Use restricted JSON Schema** (as discussed in PR #210): Like MCP, constrain JSON Schema to a useful subset for `type`, `enum`, `minimum`, `maximum`, `minLength`, `maxLength`, `pattern`, `default`, and `description`. This aligns with how Session Config Options already think about schema.
+
+2. **Support multiple input modalities**:
+   - **Simple inputs**: text, number, boolean
+   - **Selections**: select (single), multiselect (multiple) with enum-based options
+   - **Sensitive inputs**: password, URL-mode for out-of-band OAuth flows (addressing PR #330 authentication pain points)
+
+3. **Work in turn context**: Elicitation requests are triggered when a turn ends with `stopReason: "elicitation_requested"`, allowing agents to ask questions naturally within the conversation flow. Agents send elicitation requests via a separate `session/elicitation` method (following the same request/response pattern as `session/request_permission`). Unlike Session Config Options (which are persistent), elicitation requests are transient and turn-specific.
```
benbrandt: Does the agent need to return a stop_reason? For example, when permissions are requested, there isn't a stop reason, the agent may just await a response before continuing

docs/rfds/elicitation.mdx
```diff
@@ -0,0 +1,403 @@
...
+
+4. **Support client capability negotiation**: Clients declare what elicitation types they support (similar to the client capabilities pattern emerging in the protocol). Agents handle gracefully when clients don't support elicitation.
```
benbrandt: I definitely think we should follow the MCP capablity model here:
```
{
  "capabilities": {
    "elicitation": {
      "form": {},
      "url": {}
    }
  }
}
```

Where we distinguish between the two forms so we can also better map and pass this along to agents to pass to their MCP clients, and also support clients who may only be able to offer one or the other for various reasons

docs/rfds/elicitation.mdx
```diff
@@ -0,0 +1,403 @@
...
+This proposal follows MCP's established elicitation patterns. See [MCP Elicitation Specification](https://modelcontextprotocol.io/specification/2025-11-25/client/elicitation) for detailed guidance. ACP will use the same JSON Schema constraint approach, but adapted for our session/turn-based architecture.
+
+Key differences from MCP:
+- MCP elicitation is tool-call-scoped; ACP elicitation is session/turn-scoped
+- ACP must integrate with existing Session Config Options (which also use schema constraints)
+- ACP should support out-of-band flows for sensitive data (authentication from PR #330)
+
+### Elicitation Request Structure
+
+When a turn ends with `stopReason: "elicitation_requested"`, the agent sends a separate elicitation request (following the same pattern as permission requests). Example 1 (User Selection - from PR #340):
```
benbrandt: Again, I don't think the agent needs to end their turn, the client would just respond to a request, same as auth

docs/rfds/elicitation.mdx
```diff
@@ -0,0 +1,403 @@
...
+### Input Types
+
+Following MCP's approach, we would start with these types. Clients should gracefully degrade unknown types to `text`:
+
+- `text` - Open-ended text input
+- `number` - Numeric input
+- `select` - Single-choice selection from a list
+- `multiselect` - Multiple-choice selection
+- `boolean` - Yes/no choice
+- `password` - Masked text input (for sensitive credentials)
```
benbrandt: Some of these don't seem to match MCP, like this one isn't here (in fact MCP explicitly says _not_ to ask for passwords in these forms)

I also think stuff like this would be a `format` field on field of `type: string` (this is json schema after all)

benbrandt: I think the client needs to know it needs to support the restricted json schema and can decide how to represent that. We don't necessarily need to specify specific input types in the protocol definition in my opinion (again, also just looking to the MCP specification here)

docs/rfds/elicitation.mdx
```diff
@@ -0,0 +1,403 @@
...
+- `default` - Default value if user doesn't respond (agents should always provide this, even if `null`)
+- `description` - Help text explaining what's being requested
+- `enum` - Array of allowed values (for select/multiselect)
+- `minLength`, `maxLength` - String length constraints
+- `minimum`, `maximum` - Numeric range constraints
+- `pattern` - Regex pattern for validation
+
+**Not supported** (to keep initial implementation simple):
+- Complex nested objects/arrays
+- `allOf`, `anyOf`, `oneOf`
```
benbrandt: We need `oneOf` to represent single select enums: https://modelcontextprotocol.io/specification/2025-11-25/client/elicitation#requested-schema

docs/rfds/elicitation.mdx
```diff
@@ -0,0 +1,403 @@
...
+- Conditional validation
+- Custom formats
+
+This constraint list can expand in future versions based on community feedback.
+
+### Turn Response with Elicitation Stop Reason
+
+When an agent reaches a decision point and needs structured user input, it ends the turn with `stopReason: "elicitation_requested"`:
```
benbrandt: Again, this feels unnecessary in the current flow

docs/rfds/elicitation.mdx
```diff
@@ -0,0 +1,403 @@
...
+### User Response
+
+When the user responds to an elicitation request, the client sends a separate `session/elicitation` response:
+
+```json
+{
+  "jsonrpc": "2.0",
+  "id": 43,
+  "result": {
+    "elicitationResponse": {
```
benbrandt: Does this need to be nested? I think it is implied that this is an elicitation response?

docs/rfds/elicitation.mdx
```diff
@@ -0,0 +1,403 @@
...
+    "protocolVersion": "2025-11-25",
+    "clientCapabilities": {
+      "fs": {
+        "readTextFile": true,
+        "writeTextFile": true
+      },
+      "terminal": true,
+      "elicitation": {
+        "supported": true,
+        "supportedTypes": ["text", "number", "select", "multiselect", "boolean"]
```
benbrandt: again, would follow MCP's lead here so we can also use the same capabilities to declare support for MCP servers as well.

docs/rfds/elicitation.mdx
```diff
@@ -0,0 +1,403 @@
...
+  "params": {
+    "protocolVersion": "2025-11-25",
+    "clientCapabilities": {
+      "fs": {
+        "readTextFile": true,
+        "writeTextFile": true
+      },
+      "terminal": true,
+      "elicitation": {
+        "supported": true,
```
benbrandt: Supported is implied by the object being here

docs/rfds/elicitation.mdx
```diff
@@ -0,0 +1,403 @@
...
+- Clients that don't understand an elicitation type should treat it as requesting text input
+
+## Frequently asked questions
+
+### Can an agent request multiple pieces of information in one turn?
+
+For v1, we recommend a **single elicitation per turn**. This keeps the design simple and predictable for both clients and agents. It also follows the Session Config Options pattern of having agents send full state updates.
+
+If an agent needs to collect multiple pieces of information, it can:
+1. Ask one question per turn (with sensible defaults)
```
benbrandt: This shouldn't be tied to the turn lifecycle in my opinion. It is just a request that can be made

docs/rfds/elicitation.mdx
```diff
@@ -0,0 +1,403 @@
...
+
+### How does this differ from session config options?
+
+Excellent question from PR #210 discussions. Both use restricted JSON Schema, but serve different purposes:
+
+| Aspect | Session Config Options | Elicitation |
+|--------|------------------------|-------------|
+| **Lifecycle** | Persistent, pre-declared at session init | Transient, appears during turns |
+| **Scope** | Session-wide configuration | Single turn/decision point |
+| **Defaults** | Required (agents must have defaults) | Required (agents should always provide) |
```
benbrandt: Are defaults required on elicitation? I don't think so? I think the point is the json schema can allow for required fields?

docs/rfds/elicitation.mdx
```diff
@@ -0,0 +1,403 @@
...
+4. Browser returns token/credential to client (e.g., via redirect or callback)
+5. Client includes token in next `session/turn` via `elicitationResponse`
+
+**Key guarantee**: Credentials never flow through the agent or LLM, addressing the core pain point from PR #330.
+
+The exact semantics of how tokens are returned from the browser and how `returnValueFormat` is handled will be specified in detail during the implementation phase of this RFD.
+
+### Can agents use elicitation for information required before responding?
+
+Yes. An agent can include an elicitation request in a turn response with a default value and continue, then incorporate the user's response into the next turn. This is how agents can guide users through multi-step workflows.
```
benbrandt: again, I think this is where tying it to turns is an anti-pattern.
By modeling it as a request / response, the agent can decide in its own control flow whether or not to wait for a response before doing something else

docs/rfds/elicitation.mdx
```diff
@@ -0,0 +1,403 @@
...
+
+### What if a user doesn't respond to an elicitation request?
+
+The agent's default value is used (which agents must always provide). If an agent truly requires user input and wants to block, it should fail the turn and let the client handle retry logic.
```
benbrandt: I think it is fair to say that an elicitation request requires a response, even if that response is "cancelled" (we should allow for request cancellation, we can tie it together with the request cancellation changes in that RFD)

docs/rfds/elicitation.mdx
```diff
@@ -0,0 +1,403 @@
...
+
+### Should elicitation support complex nested data structures?
+
+For the initial version: no. We're focusing on simple types (strings, numbers, booleans, arrays of those). Complex nested structures can be added in future versions if use cases emerge. This keeps the initial scope manageable and lets us learn from real-world usage.
```
benbrandt: I think the key point is we are supporting whatever MCP supports here

docs/rfds/elicitation.mdx
```diff
@@ -0,0 +1,403 @@
...
+
+Agents should always design to gracefully degrade:
+- Provide sensible default values
+- Describe what they're requesting in turn content (text)
+- Proceed with the defaults
+- Clients declare capabilities so agents can make informed decisions
+
+### Can we extend this to replace the existing Permission-Request mechanism?
+
+Potentially, but that's out of scope for this RFD. PR #210 discussed that elicitation "could potentially even replace the Permission-Request mechanism" (Phil65), but that requires separate analysis of the permission request use cases and whether elicitation's constraints (no complex nesting, simpler lifecycle) are sufficient.
```
benbrandt: A point in favor of keeping these separate: since permission requests are more of a security concern, they should be handled separately so that the Client can offer a consistent experience.

Me deciding to allow a tool call should be distinct from the model asking for clarification. Also a reason that was brought up to keep the auth flow distinct. Maybe we reuse some types, but I don't think we should necessarily conflate the features

benbrandt (commented):
OK left some notes. 
I would also appreciate to have the RFD merged separately from the Rust changes. as some of the things you have added would fit better within the SDK (this crate should be mostly limited to schema types)

So before we can merge this, I would want these separated to focus the review and also save you a bunch of extra work

yordis:
> I would also appreciate to have the RFD merged separately from the Rust changes.

That will happen for sure! Please ignore for now, 🙏🏻 I will revert the code tomorrow when I wake up and move it to another branch for my own sake.

yordis:
@benbrandt to what extent I should simplify the RFD and simply point to MCP spec? Otherwise, I am getting lost in term of making sure I added all the semantic correctly. And I can't see what exactly is just an addon from ACP perspective.

Any thoughts on the topic?

docs/rfds/elicitation.mdx
```diff
@@ -0,0 +1,775 @@
...
+
+We propose introducing an elicitation mechanism for agents to request structured information from users, aligned with [MCP's draft elicitation specification](https://modelcontextprotocol.io/specification/draft/client/elicitation). This addresses discussions from PR #340 about standardizing user selection flows and PR #330 about secure authentication handling.
+
+The mechanism would:
+
+1. **Use restricted JSON Schema** (as discussed in PR #210): Like MCP, constrain JSON Schema to a useful subset—flat objects with primitive properties (`string`, `number`, `integer`, `boolean`) plus supported formats and enum values. Clients decide how to render UI based on the schema.
+
+2. **Support two elicitation modes** (following [MCP SEP-1036](https://modelcontextprotocol.io/community/seps/1036-url-mode-elicitation-for-secure-out-of-band-intera)):
+   - **Form mode** (in-band): Structured data collection via JSON Schema forms
+   - **URL mode** (out-of-band): Browser-based flows for sensitive operations like OAuth (addressing PR #330 authentication pain points)
```
josevalim, line 42: I have a question about this. Currently, elicitation is part of a session, but some agents may not allow a session to start until authentication happens. @benbrandt briefly touched on this during #330. So I am wondering if this would be suitable for OAuth.

My second concern is that some clients implements OAuth "weirdly". IIRC, Codex uses a local server, GitHub Copilot uses device OAuth which requires a code to be passed around.

Those differences make sense once you consider the use cases: agents are running locally (hence the device OAuth), MCPs often require talking to an external service. Agents require auth to start, MCPs can ask for authentication only when first invoked.

All that said, I am not against this proposal. In fact, I am quite in favor! But I do wonder if this will effectively address the OAuth concerns. In cases it is indeed unclear, we could start only with the form mode.

yordis: @josevalim I think we need to distinguish between:

1. **Agent Authentication** (what you're describing):
   - User authenticates TO the agent itself before session starts
   - Pre-session auth flow
   - Different patterns: device OAuth (Copilot), local server (Codex)
   - Blocking: session cannot start until complete

2. **Resource Authentication** (what this RFD addresses):
   - Agent requests access to third-party resources DURING session
   - In-session auth flow  
   - Example: "I need GitHub access to read your repos"
   - Non-blocking: session already running, agent needs additional capability
   
   
At least that is how I understand it. I would expect the session to exist by the time you get to this one.

josevalim: That answers it, thank you!

benbrandt: I think we likely need _some_ way to do the same sort of thing outside a session, but I also think the auth lifecycle maybe can have some special affordance for this, but re-use the same types for the request/response to make things easier for both agent/client to re-use functionality.

Still need to figure this out.. but I think we can at least unlock question asking flow from agents + mcp server things with this

benbrandt: One option would be to not have it be a `session/elicitation` method, but rather an `elicitation` method with an optional session id. If present, it is for a specific session. If not there, then it is a "global" elicitation.

Or we just go with an `authenticate/elicitation` or something

PR #377, opened 2026-01-12
RFD to rework Rust SDK, modeled on `sacp`

nikomatsakis:


benbrandt:
Happy to see this move forward! Going to accept and we can start the work. As we sort this out, I think it will be great to highlight to the other languages

PR #378, opened 2026-01-13
docs(rfd): Session Config Options to Preview stage

benbrandt:
With several implementations in place, it feels like this is ready for preview.
I added this in Zed + Codex ACP, and the JetBrains folks have this in their Kotlin SDK. So let's move this to preview stage for final feedback!

### DIFF
--- a/docs/protocol/draft/session-config-options.mdx
+++ b/docs/protocol/draft/session-config-options.mdx
@@ -123,16 +123,16 @@ Each config option **MAY** include a `category` field. Categories are semantic m
 
 <Warning>
   Categories are for UX purposes only and **MUST NOT** be required for
-  correctness. Clients **MUST** handle missing or unknown categories gracefully
-  by treating them as `other`.
+  correctness. Clients **MUST** handle missing or unknown categories gracefully.
 </Warning>
 
-| Category        | Description                                 |
-| --------------- | ------------------------------------------- |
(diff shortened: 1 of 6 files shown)


docs/protocol/draft/session-config-options.mdx
```diff
@@ -0,0 +1,281 @@
...
+  correctness. Clients **MUST** handle missing or unknown categories gracefully
+  by treating them as `other`.
+</Warning>
+
+| Category        | Description                                 |
+| --------------- | ------------------------------------------- |
+| `mode`          | Session mode selector                       |
+| `model`         | Model selector                              |
+| `thought_level` | Thought/reasoning level selector            |
+| `other`         | Unknown or uncategorized selector (default) |
```
nikomatsakis: Q: Can clients add other categories? Should we require that if clients opt for a "custom" category, it begins with `_` for extensibility purposes?

Q: Can there be *multiple* categories? (presumably not)

benbrandt: Good point. I do think this other could just be a string, with maybe some convention around _ or something

nikomatsakis: I would say something like: category names beginning with `_` are free for others to use; category names that do not begin with `_` are reserved for the ACP spec.

PR #392, opened 2026-01-21
RFD: Agent-to-Client Logging

chazcb:
## Summary

This RFD proposes a capability-gated `log` notification (agent → client) so agents can share diagnostic messages without polluting conversation history.

### Key features
- **Capability-gated**: Clients opt in via `clientCapabilities.logging`; agents only send logs to clients that declare the capability
- **Level filtering**: Clients can specify minimum log level (default: `info`)
- **Optional `sessionId`**: Connection-wide if absent, session-specific if present
- **RFC 5424 severity levels**: debug, info, notice, warning, error, critical, alert, emergency
- **Optional fields**: logger name, timestamp, structured data

### Problem

Today, agents have limited ways to inform clients about status that might impact their experience:
- **JSON-RPC errors**: Terminate the request immediately (can't use for non-fatal conditions)
- **`session/update`**: Puts diagnostics in chat history (not appropriate for transient status)

Neither works when:
- There's no active JSON RPC request to attach an error response to
- We don't want to fail the request (e.g., retries, rate limiting, fallback selection)
- There's no session yet (diagnostics after `initialize` but before `session/new`)
- We don't want to put diagnostics in chat history

### Example

```json
{
  "jsonrpc": "2.0",
  "method": "log",
  "params": {
    "level": "warning",
    "message": "Backing model rate limited, retrying in 5 seconds...",
    "sessionId": "abc-123",
    "logger": "model",
    "timestamp": "2025-01-21T10:30:00Z",
    "data": { "model": "claude-3", "retryIn": 5 }
  }
}
```

### Capability negotiation

```json
{
  "method": "initialize",
  "params": {
    "clientCapabilities": {
      "logging": {
        "level": "warning"
      }
    }
  }
}
```

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/docs/rfds/server-logging.mdx
+++ b/docs/rfds/server-logging.mdx
@@ -8,9 +8,9 @@ Author(s): [@chazcb](https://github.com/chazcb)
 
 > What are you proposing to change?
 
-Add a `log` notification that allows agents to send diagnostic messages to clients without terminating operations. This enables real-time visibility into agent health, warnings, and errors.
+Add a `log` notification that allows agents to send diagnostic messages to clients without terminating operations.
 
-**Key insight**: JSON-RPC error responses terminate the request. But many situations (MCP server disconnected, approaching rate limit, using fallback model) warrant notifying the client *without* stopping the run.
+JSON-RPC error responses terminate the request—the run is over. But many situations (approaching rate limit, using fallback model, retrying a failed request) warrant notifying the client *without* stopping the run. This proposal adds that capability.
 
 ## Status quo
 
@@ -27,9 +27,9 @@ ACP has no mechanism for agents to proactively notify clients about diagnostic i
 4. **No standard severity levels**: No way to distinguish errors from warnings from informational messages.
 
 **Real-world examples:**
-- MCP server disconnects, but other tools still work → client should know, but run continues
-- Approaching rate limit → warn the user, don't abort
-- Using fallback model → informational, definitely don't abort
+- Backing model rate limited → warn the user, retry in progress
+- Backing service disconnected → informational, reconnecting
+- Using fallback model → informational, run continues
 - Server shutting down in 5 minutes → warn all sessions
 
 ## What we propose to do about it
@@ -44,12 +44,12 @@ Add a `log` notification (agent → client) that requires capability negotiation
   "method": "log",
   "params": {
     "level": "warning",
-    "message": "MCP server 'database-tools' disconnected, retrying...",
+    "message": "Backing model rate limited, retrying in 5 seconds...",
     "sessionId": "abc-123",
-    "logger": "mcp",
+    "logger": "model",
     "code": -32001,
     "timestamp": "2025-01-21T10:30:00Z",
-    "data": { "server": "database-tools", "retryIn": 5 }
+    "data": { "model": "claude-3", "retryIn": 5 }
   }
 }
 ```
@@ -128,7 +128,7 @@ The key question is: **should the run stop?**
 - JSON-RPC error responses terminate the request—the run is over
 - `log` notifications are independent—the run continues
 
-Many situations (MCP server disconnected but others work, approaching rate limit, using fallback model) warrant notifying the client without aborting.
+Many situations (backing model rate limited, approaching context limit, using fallback model) warrant notifying the client without aborting.
 
 Also: not all agents use `prompt()`. Some use extension methods like `_enqueue` (a notification). Notifications have no response, so there's nowhere to return an error.
 
--- a/docs/rfds/server-logging.mdx
+++ b/docs/rfds/server-logging.mdx
@@ -20,7 +20,7 @@ ACP has no mechanism for agents to proactively notify clients about diagnostic i
 
 1. **JSON-RPC errors terminate requests**: If an agent returns an error, the run stops. There's no way to report non-fatal issues.
 
-2. **`session/update` requires a session**: Connection-wide errors (MCP server failures, server shutting down) can't be communicated because there's no session to attach them to.
+2. **`session/update` requires a session**: Connection-wide issues (server shutting down, configuration problems) can't be communicated because there's no session to attach them to.
 
 3. **Errors during long runs are invisible**: A prompt turn can run for minutes. If something goes wrong mid-run but the agent can recover, the client has no visibility.
 
@@ -107,7 +107,7 @@ Several reasons:
 
 3. **Opt-in presentation**: Not all clients care about logs. A separate channel lets clients ignore them entirely without polluting the conversation stream.
 
-4. **Connection-wide messages**: Errors can occur before any session exists (e.g., MCP server failing to connect after `initialize` but before `session/new`). `session/update` requires a `sessionId`, so it simply can't be used in these cases.
+4. **Connection-wide messages**: Errors can occur before any session exists (e.g., after `initialize` but before `session/new`). `session/update` requires a `sessionId`, so it simply can't be used in these cases.
 
 ### Why not use a separate channel (stderr, separate SSE endpoint, etc.)?
 
--- a/docs/rfds/server-logging.mdx
+++ b/docs/rfds/server-logging.mdx
@@ -8,314 +8,179 @@ Author(s): [@chazcb](https://github.com/chazcb)
 
 > What are you proposing to change?
 
-Add a `log` notification that allows agents to send diagnostic messages (errors, warnings, info, etc.) to clients. This notification supports both connection-wide and session-specific messages via an optional `sessionId` field.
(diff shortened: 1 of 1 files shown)


docs/rfds/server-logging.mdx
```diff
@@ -0,0 +1,178 @@
...
+
+Add a `log` JSON-RPC notification that is explicitly capability-gated. Clients opt in via `clientCapabilities.logging`; agents only send logs to clients that declare the capability. Clients can optionally specify a minimum log level.
+
+```json
+{
+  "method": "initialize",
+  "params": {
+    "clientCapabilities": {
+      "logging": {
+        "level": "warning"
```
benbrandt: I was going to say as a notification we could just add this, but having some sort of filter totally makes sense here.

Another option would be the default filter level is "error" and the client can opt-in to allowing more?

This also gives me some pause as it feels like we are mixing configuration with capabilities, so I wonder if this needs to live somewhere else, but just a gut reaction

docs/rfds/server-logging.mdx
```diff
@@ -0,0 +1,178 @@
...
+
+### Add a notification type to `session/update`
+
+Extend `session/update` with a new notification type for diagnostics. This keeps diagnostics within the existing session machinery but has drawbacks: it requires a session (can't send connection-wide diagnostics), risks polluting chat history unless clients explicitly filter, and overloads `session/update` with non-conversation concerns. Additionally, ACP specifies that session history is replayed on `session/load`, but diagnostic logs are transient and shouldn't be replayed—they're not part of the conversation.
+
+### Structured `status` notification
+
+Instead of general-purpose logging, define a more structured `status` notification explicitly for lightweight status info—similar to Claude Code's interim status messages ("Thinking...", "Searching files..."). This would be scoped to either the current session or the agent/connection level.
+
+**Tradeoffs**: More constrained semantics could be clearer for clients, but less flexible. Logging with severity levels is a well-understood pattern; inventing a new "status" abstraction may not add value over `log` with `level: info`.
```
benbrandt: Makes sense. I do feel like this more structured status would be super helpful, but understand that you are trying to get at a different use case here.

I think the clients would benefit a lot from these more interactive status messages, but you are proposing something else here

docs/rfds/server-logging.mdx
```diff
@@ -0,0 +1,178 @@
...
+
+Use HTTP headers, WebSocket ping payloads, or other transport-level channels for status.
+
+**Tradeoffs**: ACP is transport-agnostic. Relying on transport-specific mechanisms would fragment implementations and lose capability negotiation.
+
+## Shiny future
+
+> How will things play out once this feature exists?
+
+- **Clear connection feedback**: Clients can surface warnings (rate limits, retries, fallbacks) so users understand what's happening.
```
benbrandt: Regarding this: I think for clients to give a meaningful, user-facing status update, we may want to go with the more structured status approach? With the current proposal, speaking from the client perspective, I don't know that we would surface these as they would potentially be too noisy/unpredictable.

benbrandt (commented):
See my notes on status vs logging. 

I think I am trying to better understand the motivating use case. In my mind, I think we would want a more semantically meaningful status update if these will be user-facing within the main UI.

However, given that we are moving more things to remote scenarios, what this seems to show me is a way for a client to subscribe to the equivalent of stderr with a filter? Which could be helpful in debug scenarios where you need to look at the logs from the client side and don't have the luxury of having access to stderr as you would from a local process.

If we are targeting the latter, then I think this could definitely provide some value. I just think (speaking from Zed) we would likely relegate this to our ACP log/debug views that the user pulls up on-demand rather than utilize them within the main UI to avoid overly noisy logs (even with WARN/ERROR it is unclear that these should always be surfaced based on the semantics, as with many logs, they serve a different purpose potentially than a UI facing status update).

Curious to get your thoughts on all of this, and if I am completely missing the purpose of what you are proposing here.

docs/rfds/server-logging.mdx
chazcb: Hmm, agree on mixing config and capabilities. At the same time, I don't like needing to add new methods for setting logging levels, especially when you want to init a session with a specific logging level. 

For our debug logging, it's important that we can get logs immediately after initialization. Other ways to accomplish that:

1. New "setLogLevel" method that can be called _prior_ to initialization
2. New field on "params" for initialize that opts the client into logging with a specific configuration (the theory being that this is not a client "capability")
3. Or finally, if the client specifies "logging" in client capabilities, then we default to sending all log levels (for now).

Thoughts on those options?

docs/rfds/server-logging.mdx
chazcb: Yeah this proposal is really about debug and / or error logging, specifically for developers trying to get a better sense of why something went wrong or what is happening under the hood. I would imagine some UIs might provide the "show logs" pane so end users can also debug, but this is really about getting the info to the client and not about structuring dynamic UIs.

docs/rfds/server-logging.mdx
chazcb: Agree this should get removed from proposal language and proposal should focus more around logging for client visibility into internals, rather than for UIs.

docs/rfds/server-logging.mdx
benbrandt: Cool yeah then we are aligned here. Basically allows a client to offer some sort of log UI

It does now beg the question for me though :smile: :
Do we want a way for a client to be able to change the requested log-level while the connection is active? I could see it being helpful to adjust this after you have already connected, but we can also say the user needs to specify/configure this beforehand

benbrandt: (this can totally be a "we do this later" thing, just a thought I had)

docs/rfds/server-logging.mdx
benbrandt: Perfect thanks!

docs/rfds/server-logging.mdx
chazcb: I updated the proposal so clients are responsible for filtering logs, which cleaning solves this. I imagine an agent can make a session config available to change log-level if client is getting too many.

docs/rfds/server-logging.mdx
```diff
@@ -0,0 +1,185 @@
...
+> What are you proposing to improve the situation?
+
+Add a `log` JSON-RPC notification that is explicitly capability-gated. Clients opt in via `clientCapabilities.logging`; agents only send logs to clients that declare the capability.
+
+```json
+{
+  "method": "initialize",
+  "params": {
+    "clientCapabilities": {
+      "logging": {}
```
benbrandt, line 42: Thinking about this more... it seems weird to me. It seems like this is expressing that the client can log, not that is receives logs...

I wonder if an alternate approach is that the agent exposes the capability? And then we do add some sort of method for the client to opt-in to/change a level? It is off by default unless the client expresses interest?

I mean with docs, we can also do it this way, but I also feel like there needs to be _some_ standard indication of desired log level to avoid performance issues if the agent decides to send debug level logs that clog the pipe so to speak

benbrandt (commented):
One more note, otherwise I think this is pretty much ready. 
my gut feeling is clients should somehow opt-in to this. It can be across sessions/global config, that's fine, but I don't know that we should send all logs by default without the client being able to control the level filter

docs/rfds/server-logging.mdx
benbrandt: 2 is the most appealing.
I would say we could standardize on an env var but that won't work for non-local things.

So yeah I guess it would make sense to me to have this be a param on the init side to basically specify the max log level you would be willing to receive.

docs/rfds/server-logging.mdx
benbrandt: Given this is mostly for debug purposes, I think just having the client set a min log level at start seems fine

benbrandt:
Circling back here @chazcb curious how you feel about this one in relation to https://agentclientprotocol.com/rfds/agent-telemetry-export

I like the goal you have here, and it is lower-effort than setting up a full OTEL pipeline... but curious if the overlap here is worth choosing one direction, or allow for these more dev facing logs to show up in a nice way on the client

chazcb:
> Circling back here @chazcb curious how you feel about this one in relation to https://agentclientprotocol.com/rfds/agent-telemetry-export
> 
> I like the goal you have here, and it is lower-effort than setting up a full OTEL pipeline... but curious if the overlap here is worth choosing one direction, or allow for these more dev facing logs to show up in a nice way on the client

I think without a perspective on other transports (websockets / sse) it might be hard to use the otel proposal. I do think a simple "log" is low lift and straightforward, but agree it overlaps in many ways with the otel proposal. Also agree otel can be hard to set up, but for sure it would be useful to have. Maybe we can think about how the otel proposal can address other transports?

PR #393, opened 2026-01-22
RFD: Session Input Options

chazcb:
## Summary

Proposal for allowing Agents to declare input options they accept for `session/new` and `session/load` requests, advertised in `InitializeResult` capabilities.

Uses MCP elicitation's JSON Schema format for defining input schemas, providing ecosystem consistency.

## Key Design Decisions

- **Discovery at init time**: Input schemas are advertised in `InitializeResult.capabilities.sessions`, enabling clients to show configuration UI before session creation
- **Elicitation schema format**: Uses the same JSON Schema subset as MCP elicitation for type definitions, encouraging ecosystem alignment
- **Separate schemas**: Distinct schemas for `newSession` vs `loadSession` operations
- **All options optional**: Agents must provide sensible defaults

## Why not pure elicitation?

We considered using MCP elicitation directly (server requests input after `session/new`), but this has UX tradeoffs:
- Session enters "pending" state during elicitation
- Clients can't show configuration UI proactively
- Discovery requires starting a session first

By advertising schemas at init time, clients can build configuration UIs that feel native rather than reactive.

## Related

- Uses schema format from [MCP Elicitation](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/376)
- Complements Session Config Options (runtime configuration during a session)

### DIFF
--- a/docs/rfds/session-input-options.mdx
+++ b/docs/rfds/session-input-options.mdx
@@ -195,13 +195,13 @@ Use the elicitation mechanism from [MCP Elicitation](https://github.com/modelcon
 
 - No new protocol additions - reuses elicitation
 - Consistent with MCP patterns
-- Dynamic - agent can ask different questions per session
+- Dynamic - config options can vary per `session/new` or `session/load` call, not bound to the connection
 
 **Cons:**
 
 - **Client cannot show config UI until session exists** - Creates awkward UX flows where clients must create a session just to discover if there are config options to show.
 - Session exists in "pending config" state - semantically awkward
-- **Agent initialization complexity** - Agent cannot directly use config for session initialization (model selection, system prompt setup, etc.) since the session is already created.
+- **Agent initialization complexity** - Agent cannot directly use config for session initialization (model selection, system prompt setup, etc.) since the session is already created. A workaround is to create a lightweight session that hands off to a configured backend session, but this adds complexity for agent developers and requires the ability to mint session IDs upfront.
 
 **Why not chosen:** This approach works for mid-session information gathering but doesn't solve the initialization use case well.
 


chazcb:
Planning to update to:
1. Ensure that enums ("select") matches the select from Session Config Options https://github.com/agentclientprotocol/agent-client-protocol/pull/411
2. Change "hint" to "name" for consistency with Session Config Options
3. Update to clarify that there are two proposals in here: 1 for the richer options schema and 1 for session/new session/load config options

I might split the two proposals out in the near future depending on need!

docs/rfds/session-input-options.mdx
```diff
@@ -0,0 +1,461 @@
...
+```json
+{
+  "jsonrpc": "2.0",
+  "id": 1,
+  "result": {
+    "protocolVersion": 1,
+    "agentCapabilities": { ... },
+    "agentInfo": { ... },
+    "configOptions": {
+      "session/new": [
```
benbrandt: I wonder if we can somehow come up with a more structured way of handling this since we know which resources this affects, but maybe it is also fine to tie this to method names since those are stable enough. Just recording my indecision 😄

docs/rfds/session-input-options.mdx
```diff
@@ -0,0 +1,461 @@
...
+### Option types
+
+#### Select (single choice from a list)
+
+```typescript
+interface SelectOption extends ConfigOptionBase {
+  type: "select";
+  defaultValue?: string | number | boolean | null;
+  options: Array<{
+    value: string | number | boolean | null; // Machine-readable value
```
benbrandt: Question: why allow for multiple value types here and not just strings?

docs/rfds/session-input-options.mdx
```diff
@@ -0,0 +1,461 @@
...
+```
+
+### Option types
+
+#### Select (single choice from a list)
+
+```typescript
+interface SelectOption extends ConfigOptionBase {
+  type: "select";
+  defaultValue?: string | number | boolean | null;
```
benbrandt: Does this being optional indicate that this is a required field if no default is provided?

docs/rfds/session-input-options.mdx
```diff
@@ -0,0 +1,461 @@
...
+```
+
+#### Multi-select (multiple choices from a list)
+
+```typescript
+interface MultiSelectOption extends ConfigOptionBase {
+  type: "multiselect";
+  defaultValue?: Array<string | number | boolean | null>;
+  options: Array<{
+    value: string | number | boolean | null;
```
benbrandt: Same value question here

docs/rfds/session-input-options.mdx
chazcb: 100% agree 😂

docs/rfds/session-input-options.mdx
chazcb: No, the intention is that none of these fields are required. This `defaultValue` is intended to be a hint to the client about what the agent _might_ use if the field is not filled out.

docs/rfds/session-input-options.mdx
chazcb: This is based on our internal simplified JSON schema. I didn't realize that elicitation also uses a simplified schema, so love the idea of moving to that instead.

benbrandt: cool yeah I think it will allow clients and agents to reuse UI + validation logic for this stuff if we can keep it closer, and I suspect it will likely cover what we need here

docs/rfds/session-input-options.mdx
benbrandt: Got it. This will be one difference from the elicitation json schema which is it will not be allowed to have "required" fields. 
But I think we can learn a lot from this and potentially allow required fields later if it feels useful

docs/rfds/session-input-options.mdx
chazcb: Updated to the MCP json schema

docs/rfds/session-input-options.mdx
chazcb: Update to note that if the client sends _any_ part of the schema they should send the whole part, that way there is no confusion and agent can expect either the default or the full config.

docs/rfds/session-input-options.mdx
chazcb: Updated to the MCP elicitation json schema

docs/rfds/session-input-options.mdx
chazcb: Maybe a better option is to have a discovery method like "session/new_options" and server returns the config.

docs/rfds/session-input-options.mdx
```diff
@@ -0,0 +1,241 @@
...
+**Pros:**
+
+- Reuses elicitation infrastructure
+- Agent receives config before session creation - can use for initialization decisions
+- Single mechanism for all elicitation needs
+
+**Cons:**
+
+- **Client still cannot show config UI before user clicks "New Session"** - Same UX limitation as Alternative A.
+- **Correlation problem** - Without a session ID, client doesn't know which pending `session/new` request the elicitation belongs to. Requires protocol additions (e.g., `relatedRequestId`) to link them.
```
anna239, line 231: we're currently working on ability to provide requestId in the elicitation request — https://github.com/agentclientprotocol/agent-client-protocol/pull/792 We need this functionality for things other then config options, will it maybe work for this scenario too?

eamarucci:
### Contextual pre-session model discovery, including startup and authentication guarantees

Following the [suggestion of a discovery method](https://github.com/agentclientprotocol/agent-client-protocol/pull/393#discussion_r2774713046), could this RFD cover a supported, negotiated contextual query for models and per-model reasoning options before creating a session, including an explicit refresh? A launcher needs choices for its intended working directory and provider/configuration context before the first launch. A connection-wide schema alone may not capture those differences.

In a static review of codex-acp at `acc035a7444bf7550aabcc24597ed845ba3e593b`, [newSession creates a thread before fetching models](https://github.com/agentclientprotocol/codex-acp/blob/acc035a7444bf7550aabcc24597ed845ba3e593b/src/CodexAcpClient.ts#L599-L619). The adapter wraps App Server model/config reads, but that does not establish that a pre-session query would use the same context as a later launch. I have not run a reproduction.

Could the supported discovery path cover startup and initialize as well as the query, without creating or resuming an operational session, sending a prompt, running inference, mutating authentication, or exposing credentials? The current [launcher starts App Server before ACP initialize](https://github.com/agentclientprotocol/codex-acp/blob/acc035a7444bf7550aabcc24597ed845ba3e593b/src/index.ts#L78-L149), and [initialize schedules an account-status read](https://github.com/agentclientprotocol/codex-acp/blob/acc035a7444bf7550aabcc24597ed845ba3e593b/src/CodexAcpServer.ts#L1173-L1223). Passing `refreshToken:false` to one read does not establish that the whole startup path preserves authentication. If these constraints cannot be met for a context, an explicit unavailable result would be useful. This is not a blanket objection to an internal ephemeral process or thread; its effects and lifecycle would need to satisfy the same constraints.

Partial information should remain explicit: unsupported discovery, an unavailable or empty catalog, and unqueried model-specific options should not become invented defaults or evidence that a model can execute. A default may be omitted; reasoning options need to remain associated with their model. Cancellation and resource-closure semantics should be documented, with process-exit observation remaining the client/supervisor’s responsibility.

Is this contextual query appropriate for this RFD, or should the adapter first define an optional extension? I am asking for a supported interface and clarification of its guarantees, without prescribing a method name, response schema, or timeout values.

benbrandt:
I think we've discussed this a bit, and a better approach will probably be to have a `session/config` or something similar that is a request that takes a cwd, addl directories, etc and returns which config options would be available on a new/loaded session, so that you can pass them in to the session creation.

Since many of these values can resolve based on project configs, I think having the response be based on similar config as a session would make a lot of sense, but would also allow the client to know what would be available beforehand

PR #395, opened 2026-01-24
RFD: Session Delete

chazcb:
## Summary

- Proposes a capability-gated `session/delete` method to complement `session/list`
- Gives users control over which sessions appear in their session history
- Implementation-defined storage behavior (soft/hard delete) and load behavior for deleted sessions

Follows up on the [session/list RFD](https://agentclientprotocol.com/rfds/session-list#what-about-session-deletion) which explicitly deferred deletion.

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/docs/docs.json
+++ b/docs/docs.json
@@ -119,7 +119,8 @@
               "rfds/acp-agent-registry",
               "rfds/auth-methods",
               "rfds/rust-sdk-v1",
-              "rfds/logout-method"
+              "rfds/logout-method",
+              "rfds/session-delete"
             ]
           },
           { "group": "Preview", "pages": ["rfds/session-config-options"] },
--- a/docs/updates.mdx
+++ b/docs/updates.mdx
@@ -5,6 +5,13 @@ rss: true
 ---
 
 <Update label="February 4, 2026" tags={["RFD"]}>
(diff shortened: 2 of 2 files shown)


docs/rfds/session-delete.mdx
```diff
@@ -0,0 +1,138 @@
...
+
+> What are you proposing to improve the situation?
+
+Add a `session/delete` JSON-RPC method that is capability-gated. Clients opt in via `sessionCapabilities.delete`; agents advertise this capability at initialization if they support deletion.
+
+```json
+{
+  "method": "initialize",
+  "params": {
+    "clientCapabilities": {
```
benbrandt: Hmm wouldn't this be an agent capability? And agents can offer the option to delete if they want based on this?

benbrandt (changes requested):
One small note, otherwise I think looks good and we can move this one forward.

docs/rfds/session-delete.mdx
chazcb: Hah yup. Updating now.

chazcb: Fixed this, thanks for noting!

PR #411, opened 2026-01-27
feat: Stabilize Session Config Options

benbrandt:
Proposing to move this to the Completed stage!

### DIFF
--- a/docs/updates.mdx
+++ b/docs/updates.mdx
@@ -4,6 +4,13 @@ description: Updates and announcements about the Agent Client Protocol
 rss: true
 ---
 
+<Update label="February 4, 2025" tags={["RFD"]}>
+## Session Config Options RFD moves to Completed
+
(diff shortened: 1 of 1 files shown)


chazcb:
@benbrandt would love if you took a quick look at https://github.com/agentclientprotocol/agent-client-protocol/pull/393 and the different config schema there. It might make sense to consolidate the way we define config schemas together.

benbrandt:
@chazcb a few thoughts:

- First: sorry for not doing a full review on that one sooner
- Second: we made session config options possible to add new variants in the future, so I don't think it is at odds
- Third: Session config options has something yours doesn't yet, which is the ability to have a distinction between values and labels, i.e. a unique ID and a user-facing value, but we can address that on your RFD

benbrandt:
OK going to stabilize this!

PR #441, opened 2026-02-02
docs(rfd): Indicate deletes in diffs

anna239:


docs/rfds/diff-delete.mdx
```diff
@@ -0,0 +1,57 @@
...
+
+## Shiny future
+
+> How will things will play out once this feature exists?
+
+It is possible for the agent to distinguish between a deleted file and an empty file.
+
+## Implementation details and plan
+
+> Tell me more about your implementation. What is your detailed implementation plan?
```
yordis, line 58: Could you add an example message representing the usage? 🙏🏻

yordis (commented):
My primary concern are,

## Boolean explosion

If we add `deleted`, Would we add `created`, `renamed`, etc? Are we comfortable turning diff into a growing bag of flags with unclear combinations?

## Explicit operation

Isn’t `deleted` an operation (the file’s post-state) rather than an attribute? Wouldn’t a single kind: `added|modified|deleted` be the more direct representation and closer to how tools like git model it?

## Semantic complexity

What does `deleted=true` actually mean for newText and validation rules? Must `newText` be empty, ignored, or still applied? How do we prevent contradictory payloads like `deleted=true` with non-empty `newText`, or `deleted=false` with `/dev/null-style` semantics later?

benbrandt (commented):
@anna239 let's turn this into the long-awaited new version of diff that supports more info so we can represent stuff like add/delete/move/edit and maybe include stuff like line numbers etc

ignatov:
I think that deleted here is only for the sake of compatibility, as we don't want to have a breaking change for a tiny fix

ignatov:
> @anna239 let's turn this into the long-awaited new version of diff that supports more info so we can represent stuff like add/delete/move/edit and maybe include stuff like line numbers etc

@benbrandt I suppose the can do a tiny non-breaking change and after do some real fixes in the product, after we can start a new wave of discussions towards our shiny feature

anna239:
@benbrandt I don't think there are any. In future if we go with introducing new diff format, it would be nice to support 'file move', but currently it's possible to mimic that with file removed - file added logic

PR #442, opened 2026-02-02
Added remote-session-support.mdx.mdx

anna239:


anna239:
@benbrandt I thought about this and I think we won't be able to abstract away the fact that agent execution doesn't happen on the current machine and that the changed code is not available right away. So I would prefer to keep it more explicit in the protocol

docs/rfds/remote-session-support.mdx
```diff
@@ -0,0 +1,167 @@
...
+- Connection management and lifecycle
+- Server-side implementation details
+
+These concerns will be addressed in separate RFDs.
+
+## Status quo
+
+> How do things work today and what problems does this cause? Why would we change things?
+
+Currently, ACP sessions assume the agent runs locally alongside the client, with direct filesystem access. The agent can read and write files in the user's working directory.
```
benbrandt, line 30: If I understand correctly: this is more about which code the agent should be working on?
Is the code in the `cwd`, or will it have to fetch it?

In this scenario, is it important that the agent knows what to do with this information? Or will there be something in between?

If the agent should, then I guess we need some capabilities. If not, then I wonder if this is somehow better handled by the proxy RFD where something gets a cwd ready for the agent session before calling session/new for the agent?

benbrandt: Basically, I am trying to understand whose responsibility it is to set this thing up which will help me better grasp how this should fit in the protocol

benbrandt: I.e. should this be here? Another method? somewhere else? etc.

alpgul:
I would like to make a specific suggestion for this PR. To eliminate the agent's dependency on a `workdir`, we can move tool management to the client-side.

If we define the MCP (Model Context Protocol) structures on the client and bind them with the agent, the agent can trigger these tools through the client whenever it needs to perform an action. This removes the necessity of hosting the entire project on a remote server, allowing the agent to function in a completely stateless manner, focusing solely on reasoning. This architecture could be more efficient in terms of both data privacy and performance.

anna239:
> If we define the MCP (Model Context Protocol) structures on the client and bind them with the agent, the agent can trigger these tools through the client whenever it needs to perform an action. 

As I see it, the main advantages of remote agent execution are: 
- agent doesn't make changes in your local copy while working, you can do some other things there 
- agent can run for a long time, you don't need to keep your machine running 

Both both these ^^ advantages will be lost if tool calling happens on a client

benbrandt:
Going to go ahead and close this one. I think this feeds into the "orchestrator" concept we still need to figure out, i.e. an entity that can setup agents, rather than requiring every agent to implement this.

PR #447, opened 2026-02-02
Add account-status-method.mdx

anna239:


docs/rfds/account-status-method.mdx
```diff
@@ -0,0 +1,435 @@
...
+#### AccountStatusResponse
+
+All fields are nullable to accommodate different providers that may not support all metrics.
+
+```typescript
+interface AccountStatusResponse {
+  /** Extension metadata */
+  _meta?: Record<string, unknown>;
+
+  /** Credit/balance information */
```
yordis, line 62: From the specification perspective, having all these values feel very specific to a given product mindset.

I wonder, should the spec focus on the messaging spec and allow specific clients to start with `_meta` as a way to send the information they want?

docs/rfds/account-status-method.mdx
```diff
@@ -0,0 +1,435 @@
...
+### Supporting Types
+
+#### CreditsInfo
+
+Information about prepaid credits or account balance.
+
+```typescript
+interface CreditsInfo {
+  /** Remaining credit balance in smallest currency unit (e.g., cents) */
+  remaining?: number | null;
```
josevalim, line 85: How are we going to know how to format it? Some currencies do not have cents, so we would need to import a CLDR library to know exactly how to format them? Or should this be a string? Same as the field below.

docs/rfds/account-status-method.mdx
```diff
@@ -0,0 +1,435 @@
...
+
+  /** Total credits purchased/allocated */
+  total?: number | null;
+
+  /** Currency code (ISO 4217, e.g., "USD") */
+  currency?: string | null;
```
josevalim, line 91: For GitHub, in particular, they do not have currency, but "premium requests", so currency wouldn't work for them ([this same issue exists on SessionUsage too](https://github.com/agentclientprotocol/agent-client-protocol/pull/454)).

docs/rfds/account-status-method.mdx
```diff
@@ -0,0 +1,435 @@
...
+  inputTokens?: number | null;
+
+  /** Total output tokens used */
+  outputTokens?: number | null;
+
+  /** Total requests made */
+  requests?: number | null;
+
+  /** Total cost incurred in smallest currency unit */
+  cost?: number | null;
```
josevalim, line 171: The SessionUsage RFD calls this "amount", perhaps worth aligning?

anna239:
I've looked more into how different agents present your costs and it looks like most of them can only share data about the current session. For the general information they recommend checking account information on the website. This means even if we go forward with this rfd there will be little adoption among the agents. So I'm closing it for now

PR #455, opened 2026-02-04
docs(rfd): Move registry RFD to preview

benbrandt:
Updates the rfd to match the current implementation.


benbrandt:
@ignatov I updated this based on the docs you currently have in the registry but I'd love you to take a look. 
@nikomatsakis before we move to Preview, I'd love to have an idea of what we want to call `extensions` or remove it for now from the RFD/registry spec

ignatov:
we can remove extensions for now, I don't mind, while they're empty, it's not a big deal

PR #467, opened 2026-02-06
fix: use impl IntoOption<Meta> for CancelRequestNotification::meta()

konippi:
`CancelRequestNotification::meta()` was the only builder method across the entire codebase that accepted `Meta` directly instead of `impl IntoOption<Meta>`.


benbrandt (approved):
thank you!

PR #484, opened 2026-02-10
docs(rfd): prompt queueing RFD

SteffenDE:
A frequently requested feature for Tidewave is the ability to queue prompts like in Claude Code CLI. This is my idea of how that could work.

Now the problem with this is that I tried to find a way to actually get the Claude Agent SDK to tell me when the queued message is inserted into the context, but it looks like there is no way. The message is immediately pulled from the AsyncInterable and the generator does not include any hint when to inject it into a chat view...

I guess one way to cheat would be to just insert it whenever all pending tool results are done. Maybe that's what Claude Code does.

### DIFF
--- a/docs/rfds/prompt-queueing.mdx
+++ b/docs/rfds/prompt-queueing.mdx
@@ -1,0 +1,57 @@
+---
+title: "Prompt Queueing"
+---
+
+Author(s): [SteffenDE](https://github.com/SteffenDE)
+
+## Elevator pitch
+
+> What are you proposing to change?
+
+Add a new `promptQueueing` session capability, that tells client if the agent supports sending prompts while another prompt is already being processed for that session.
+The Claude Code CLI supports writing a user message while the agent is working and injects that message at the next opportunity (i.e. after the current turn is finished).
+
+Since ACP prompts can consist of many agent turns, having an officially supported way to adjust running prompts is very useful.
+
+## Status quo
+
+> How do things work today and what problems does this cause? Why would we change things?
+
+Today, the specification does not mention what should happen when sending a concurrent prompt request.
+Clients cannot assume that this is supported, so running prompts need to be canceled when a client wants to allow users to send a new prompt while an existing one is running.
+
+## What we propose to do about it
+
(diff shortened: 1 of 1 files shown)


SteffenDE:
Ok, so for Claude we actually can find out when the SDK processed the message by passing `extraArgs: {"replay-user-messages": ""}`. Then, the SDK replays the user message (https://platform.claude.com/docs/en/agent-sdk/typescript#sdkuser-message-replay), so we can compare the message to what was sent to know when it was sent.

docs/rfds/prompt-queueing.mdx
```diff
@@ -0,0 +1,57 @@
...
+Add a new `promptQueueing` session capability, that tells client if the agent supports sending prompts while another prompt is already being processed for that session.
+The Claude Code CLI supports writing a user message while the agent is working and injects that message at the next opportunity (i.e. after the current turn is finished).
+
+Since ACP prompts can consist of many agent turns, having an officially supported way to adjust running prompts is very useful.
+
+## Status quo
+
+> How do things work today and what problems does this cause? Why would we change things?
+
+Today, the specification does not mention what should happen when sending a concurrent prompt request.
```
benbrandt, line 20: A few thoughts on this that make it tricky:

- There is a common UX to be able to edit a queued message, how do we handle this? 
- what does the response on of the /prompt request "respond" to? the second, queued message?

As an FYI, I am planning to propose soon a completely notification-based turn lifecycle (bidirectional stream of events from both sides updating the session) that will hopefully make it easier to model this sort of thing

SteffenDE: I am happy to wait for a better proposal. This is more of a band-aid solution for sure. It does not allow editing a sent message in any way. Maybe queued is actually the wrong word, since as soon as the prompt is sent, it is assumed to be processed by the agent in some way.

There are two responses:

1. The initial prompt responds as soon as the second prompt enters the agent context.
2. The second prompt responds when it is done as usual.

SteffenDE:
Closing in favor of #1261.

PR #533, opened 2026-02-18
RFD: Multi-Client Session Attach

joaommartins:
## Summary

This RFD proposes a `session/attach` method that allows multiple ACP clients to connect to and interact with the same live agent session simultaneously — enabling a unified notification and approval UI across concurrent coding agent workflows.

### Key features

- **Controller/observer roles** — Controllers can send prompts and respond to permissions; observers receive the live stream read-only
- **First-writer-wins permission routing** — `request_permission` is broadcast to all controllers; the first response wins and all clients are notified via `permission_resolved`
- **`historyPolicy` parameter** — Clients control history replay on attach: `full` (default), `pending_only`, or `none`
- **Proxy-based architecture** — A proxy/multiplexer sits between agents and clients, meaning existing agents require zero changes
- **WebSocket + SSE transport** — Multi-client attach requires a network-capable transport; the proxy handles this
- **Graceful detach and lifecycle** — `session/detach` lets clients leave voluntarily; the session continues as long as one controller remains

### Problem

ACP currently assumes a 1:1 relationship between a client and an agent session. For developers running multiple concurrent agent sessions (Claude Code, Codex, Gemini, etc.), this creates a significant context-switching tax:

- Permission requests block in whichever terminal started the session — there is no way to respond from a unified dashboard
- There is no way to passively observe an active session from a second client
- `session/load` supports sequential handoff only — not concurrent access to a live session

### Example

```json
{
  "jsonrpc": "2.0",
  "id": 5,
  "method": "session/attach",
  "params": {
    "sessionId": "sess_abc123def456",
    "role": "controller",
    "historyPolicy": "pending_only",
    "clientInfo": {
      "name": "notification-dashboard",
      "version": "1.0.0"
    }
  }
}
```

### Capability negotiation

```json
{
  "agentCapabilities": {
    "sessionCapabilities": {
      "attach": {
        "roles": ["controller", "observer"]
      }
    }
  }
}
```

### Relationship to existing RFDs

This proposal builds directly on the [Agent Extensions via ACP Proxies](./proxy-chains) RFD — the proxy architecture established there is the recommended implementation pattern for multi-client attach. It also depends on [Session List](./session-list) for session discovery and complements [Resuming of existing sessions](./session-resume).

### DIFF
--- a/docs/rfds/multi-client-session-attach.mdx
+++ b/docs/rfds/multi-client-session-attach.mdx
@@ -1,0 +1,352 @@
+---
+title: "Multi-Client Session Attach"
+---
+
+- Author(s): [@joaommartins](https://github.com/joaommartins)
+- Champion: <!-- to be assigned -->
+
+## Elevator pitch
+
+Allow multiple ACP clients to connect to and interact with the same live agent session simultaneously, enabling a unified notification and approval UI across concurrent coding agent workflows. A developer running several agent sessions (e.g., Claude Code, Codex, Gemini) could respond to permission requests, answer agent questions, and monitor progress from a single dashboard client — even if the sessions were started from different frontends.
+
+## Status quo
+
+ACP currently assumes a **1:1 relationship** between a client and an agent session. The transport layer (stdio) is inherently single-client: one process pipe per connection, no shared state between clients. This creates several problems for developers running multiple agent sessions:
+
+1. **No unified oversight** — Each agent session is bound to the frontend that started it. If you start Claude Code in terminal A, you can only respond to its permission requests from terminal A.
+2. **Permission requests go unanswered** — When an agent blocks on `request_permission`, the developer must find the correct terminal/tab/window to respond. With multiple concurrent sessions, this becomes a significant context-switching tax.
+3. **No observation without control** — There is no way to passively monitor an active session from a secondary client (e.g., a mobile device, a web dashboard, or a notification center).
+4. **Session handoff is sequential, not concurrent** — `session/load` allows a new client to resume a previous session, but the original client must have disconnected first. There is no mechanism for a second client to "attach" to a live, in-progress session.
+
+### Current workarounds
+
+- **Toad** and **Agentastic.dev** solve part of this by being the single frontend that starts all sessions — but you must commit to one UI upfront.
+- **`ai-agents-notifier`** and **`anot`** provide desktop notifications for permission requests but are one-way (notify only, no reply).
+- **`tmux`/`screen`** can share a terminal session, but this shares the entire terminal, not the structured ACP session.
+
+### Related RFDs
+
+This proposal builds on and intersects with several existing draft RFDs:
+
+- **[Agent Extensions via ACP Proxies](./proxy-chains)** — The proxy/multiplexer architecture pattern is central to the recommended implementation approach for multi-client sessions
+- **[Session List](./session-list)** — Discovering existing sessions (prerequisite for attach)
+- **[Resuming of existing sessions](./session-resume)** (`session/resume`) — Sequential handoff between clients
+- **Session Info Update** — Real-time metadata propagation (draft)
+- **Session Fork** — Branching sessions (related but distinct — fork creates a copy, attach shares the original) (draft)
+
+## What we propose to do about it
+
+### Core concept: Session Attach
+
+Add a `session/attach` method that allows a client to connect to an **active, in-progress session** owned by another client. The attaching client receives:
+
+1. A replay of conversation history (like `session/load`)
+2. A live stream of `session/update` notifications going forward
+3. The ability to respond to `request_permission` prompts
+4. Optionally, the ability to send `session/prompt` messages
+
+### Capability negotiation
+
+Agents (or proxies) advertise multi-client support during initialization:
+
+```json
+{
+  "agentCapabilities": {
+    "sessionCapabilities": {
+      "attach": {
+        "maxClients": 3,
+        "roles": ["controller", "observer"]
+      }
+    }
+  }
+}
+```
+
+### Client roles
+
+- **Controller** — Can send prompts, respond to permission requests, cancel operations. The original client is always a controller.
+- **Observer** — Receives `session/update` notifications and can see pending `request_permission` prompts, but cannot act on them.
+
+A client specifies its requested role when attaching:
+
+```json
+{
+  "jsonrpc": "2.0",
+  "id": 5,
+  "method": "session/attach",
+  "params": {
+    "sessionId": "sess_abc123def456",
+    "role": "controller",
+    "clientInfo": {
+      "name": "notification-dashboard",
+      "version": "1.0.0"
+    }
+  }
+}
+```
+
+The response follows standard ACP patterns, returning session metadata:
+
+```json
+{
+  "jsonrpc": "2.0",
+  "id": 5,
+  "result": {
+    "sessionId": "sess_abc123def456",
+    "role": "controller",
+    "history": [
+      {
+        "type": "prompt",
+        "content": "Add authentication to the API",
+        "timestamp": "2026-02-18T14:22:00Z"
+      },
+      {
+        "type": "permission_request",
+        "toolCallId": "call_abc",
+        "status": "pending"
+      }
+    ],
+    "connectedClients": [
+      {
+        "name": "Claude Code",
+        "version": "1.0.0",
+        "role": "controller"
+      }
+    ]
+  }
+}
+```
+
+### Permission request routing
+
+When an agent emits `request_permission`, it is broadcast to **all connected controller clients**. The first client to respond wins (first-writer-wins semantics). The agent then notifies all other clients that the permission was resolved:
+
+```json
+{
+  "jsonrpc": "2.0",
+  "method": "session/update",
+  "params": {
+    "sessionId": "sess_abc123def456",
+    "update": {
+      "type": "permission_resolved",
+      "toolCallId": "call_xyz",
+      "chosenOptionId": "allow_once",
+      "resolvedBy": {
+        "name": "notification-dashboard",
+        "version": "1.0.0"
+      }
+    }
+  }
+}
+```
+
+### Transport considerations
+
+The current stdio transport is inherently single-client. Multi-client attach requires a **network-capable transport**. We propose this works over:
+
+- **WebSocket** — For persistent, bidirectional connections from multiple clients
+- **HTTP + SSE** — For observer-only clients that only need to receive updates
+
+The agent (or an ACP proxy) would listen on a network port. This is consistent with the direction of the **[Proxy Chains RFD](./proxy-chains)**, where a proxy mediates between clients and agents.
+
+### Proxy architecture (recommended implementation pattern)
+
+Rather than requiring every agent to implement multi-client support natively, the recommended pattern is an **ACP proxy/multiplexer**, as described in the [Agent Extensions via ACP Proxies RFD](./proxy-chains):
+
+```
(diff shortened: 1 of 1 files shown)


ElleNajt:
I have a use case for this and found this issue because I was considering building the same thing:  

I've been building (WIP, I can't guarantee it's safe at the moment) [github.com/ElleNajt/agent-to-go](https://github.com/ElleNajt/agent-to-go) , where the idea is that I forward claude code terminal sessions via tmux/ttyd/tailscale to my phone. 

There are similar projects, but I'm trying to build something minimal that just provides easily auditable glue interacting with and and managing coding agent CLIs on the phone -- and multiplexing an ACP session between a mobile friendly ACP front end and whatever ACP front end is on the dev machine is ideal for this.

ElleNajt:
Claude and I vibed out a multiplexer here: https://github.com/ElleNajt/acp-multiplex -- the gist is that there's a single primary process that talks to the acp backend like claude-code-acp through a proxy, and that proxy spawns a socket that secondary front ends can connect to.

Here's an example where agent-shell is the primary and toad is the secondary:

image

Curious for takes on this approach -- I haven't reviewed this in depth yet (Claude's walk through is here: https://github.com/ElleNajt/acp-multiplex/blob/master/docs/walkthrough.org#L509 )... I just really want this to exist, and would love to collaborate with someone on manifesting a stable and safe way to mirror acp sessions on my phone. :)

ElleNajt:
Following up to say the multi-plexer is working nicely for my use case of mirroring my dev conversations to my phone. 

The only thing I've needed to add to the protocol so far is a turn_complete message, so that the agent-shell.el front end can handle conversations on the acp-mobile side. 

Permission handling across devices works, as does replay. I didn't build out anything having to do with observability metadata; every connected client is a controller. 

I would love to see the protocol officially support multiplexing, since then I wouldn't need to maintain my patch to agent-shell.el for handling turn_complete. :)

docs/rfds/multi-client-session-attach.mdx
```diff
@@ -0,0 +1,414 @@
...
+    "historyPolicy": "full",
+    "clientInfo": {
+      "name": "notification-dashboard",
+      "version": "1.0.0"
+    }
+  }
+}
+```
+
+The `historyPolicy` parameter controls what history is replayed on attach:
```
denofevil, line 75: What do you think about supporting loading the history from an arbitrary message (in regards to https://github.com/agentclientprotocol/agent-client-protocol/pull/536)? This will allow reconnecting clients to load only delta since the last message they already received

denofevil:
:+1: to having turn_complete multiplexed, but I think we also need to echo the prompt sent by the client to all other clients, otherwise they won't be able to see it and will only get session_update corresponding to that prompt

joaommartins:
Thanks for the feedback @ElleNajt and @denofevil — really appreciate it, especially the real-world validation from acp-multiplex. Great to see this working in practice already.

I've pushed an update addressing all three points:

- **`after_message` history policy** — new `historyPolicy: "after_message"` with an `afterMessageId` field, so reconnecting clients can request just the delta since their last known message. Falls back to full replay if the proxy doesn't recognise the ID. This depends on the Message ID RFD (#536) as @denofevil suggested.

- **`turn_complete` notification** — the proxy now broadcasts a `turn_complete` session update to all connected clients when the agent finishes a turn. The prompting client already gets this via the `session/prompt` response, but every other client needs an explicit signal. This should cover the gap @ElleNajt hit with agent-shell.el.

- **`prompt_received` echoing** — when a controller sends a prompt, the proxy echoes it to all other clients as a `prompt_received` session update *before* the agent starts streaming. Includes a `sentBy` field so clients know who sent it. As @denofevil pointed out, without this secondary clients would only see the response with no context.

Also updated the proxy responsibilities, implementation plan, and related RFDs section to reference Message ID.

docs/rfds/multi-client-session-attach.mdx
```diff
@@ -0,0 +1,495 @@
...
+  "jsonrpc": "2.0",
+  "id": 5,
+  "method": "session/attach",
+  "params": {
+    "sessionId": "sess_abc123def456",
+    "role": "controller",
+    "historyPolicy": "after_message",
+    "afterMessageId": "ea87d0e7-beb8-484a-a404-94a30b78a5a8",
+    "clientInfo": {
+      "name": "notification-dashboard",
```
denofevil, line 95: What do you think about providing self-assigned id along with the number? This way clients with the same name (e.g. two mobile ACP clients) can distinguish each other

docs/rfds/multi-client-session-attach.mdx
```diff
@@ -0,0 +1,495 @@
...
+    "sessionId": "sess_abc123def456",
+    "role": "controller",
+    "history": [
+      {
+        "type": "prompt",
+        "content": "Add authentication to the API",
+        "timestamp": "2026-02-18T14:22:00Z"
+      },
+      {
+        "type": "permission_request",
```
denofevil, line 119: This part is handy for notification services, so they can understand that there are missing requests and send notifications accordingly. However, it's unclear what should happen with full clients: how do they respond to pending requests? Currently, such a request is an RPC call from the agent to the client. Should the agent emit missing RPC calls for the client, or should the client respond in some other way?

docs/rfds/multi-client-session-attach.mdx
joaommartins: Done — added an optional `clientId` field to the attach request. It's a self-assigned opaque string (e.g. UUID) that the proxy propagates through `connectedClients`, `resolvedBy`, `sentBy`, and lifecycle notifications. If omitted, the proxy MAY assign one and return it in the response.

docs/rfds/multi-client-session-attach.mdx
joaommartins: Yeah this needed spelling out — added a "Pending permission replay for attaching controllers" section. The gist: history entries are display-only (so the client has context), and separately the proxy re-issues any unresolved `request_permission` calls as normal JSON-RPC requests after the attach response. From the client's POV they look like fresh permission requests, so no special handling needed. First-writer-wins still applies — if someone else resolves it first, the proxy discards the late response and sends `permission_resolved`.

denofevil (approved):

anna239:
hi @joaommartins I have a minor question — could you please elaborate why do you need two separate roles - controller and observer? If some service doesn't want to handle permission requests then could just ignore it, or am I missing something?

joaommartins:
Hi @anna239 , you're not missing anything, I just prefer an explicit client declaration. Not much is lost by not having it, except for the server submitting `request_permission` notifications to clients which do not intend to respond at all, which is a perfectly fine approach. Should I introduce this change?

anna239:
@joaommartins if we add the roles, the for each new event\notification we'll need to decide if it should be sent to observer or not. For instance I'm working on elicitation mechanism at the moment - I suppose elicitation requests shouldn't be sent to observer? I would prefer that we do not have to somehow filter events that go to different connected clients. 
Otherwise proposal looks really good to me

joaommartins:
@anna239 Yeah that's a fair call — I've removed the role distinction entirely. The proxy now just broadcasts everything to every client, and clients that don't care about certain events (permissions, elicitation, whatever comes next) simply ignore them. No per-event routing decisions, no need to revisit filtering logic as the protocol grows.

Also dropped capability negotiation from agentCapabilities since it's the proxy implementing this, not the agent — felt unnecessary to advertise it there. Clients discover attach support by calling session/attach and getting an error if unsupported.

smagnuso:
I implemented this RFD end-to-end in [hydra-acp](https://github.com/smagnuso/hydra-acp) (a proxy/daemon that multiplexes editors, Slack bridges, a web UI, and headless approver/notifier extensions onto a single agent). Five observations below from doing the work — one shape proposal, the rest small clarifications. Each is standalone; happy to PR fixes against the RFD doc.

### 1. `outcome` shape for `permission_resolved` (replaces bare `chosenOptionId`)

The current `permission_resolved` shape carries `chosenOptionId: string`, which can only represent a successful selection. It can't express the cancellation outcomes that proxies routinely need to broadcast:

- The resolving client returned `{ outcome: { kind: "cancelled" } }` because the user dismissed the modal.
- The proxy itself synthesizes a cancellation — e.g. when a session daemon restart kills in-flight `request_permission` promises, the proxy needs to tell every attached client "this is dead, dismiss your UI" with an explanatory reason.

Existing ACP `session/request_permission` responses already use `outcome: { kind, optionId? }`, so reusing that shape on the resolve event aligns the two sides of the same logical exchange:

```jsonc
{
  "sessionUpdate": "permission_resolved",
  "toolCallId": "tc_42",
  "outcome": {
    "kind": "selected",              // | "cancelled"
    "optionId": "allow_once",        // present when kind === "selected"
    "reason": "daemon-disconnected"  // optional context for kind === "cancelled"
  },
  "resolvedBy": { "clientId": "...", "name": "...", "version": "..." }
}
```

Backwards-compat is cheap — implementations can keep emitting `chosenOptionId` alongside `outcome` (redundant when `outcome.kind === "selected"`). I'm doing exactly that in hydra-acp today: both fields land on the wire, spec-strict consumers read `chosenOptionId`, outcome-aware consumers read `outcome`.

### 2. `sessionUpdate` vs `type` discriminator on the four new `session/update` variants

Minor but worth catching before this lands. The RFD examples for `permission_resolved`, `prompt_received`, `turn_complete`, and `client_disconnected` all use `type: "<name>"` as the discriminator inside `update`:

```jsonc
"update": {
  "type": "permission_resolved",
  ...
}
```

Every existing ACP `session/update` variant (`agent_message_chunk`, `tool_call`, `tool_call_update`, `current_model_update`, `available_commands_update`, `usage_update`, `session_info_update`, `plan`, …) uses `sessionUpdate: "<name>"` as the discriminator. If `type` is intentional here it'll be a meaningful break from the rest of ACP; if it's a doc typo, easier to fix now. My implementation uses `sessionUpdate` to match the established convention.

### 3. Generalize `messageId` to every recordable `session/update` for finer `after_message` resync

The RFD scopes `messageId` to `prompt_received` (id of the user prompt) and `turn_complete` (id of the final agent message). In practice, this means `after_message` reconnects can only resync at turn boundaries — a client that missed five tool calls and two plan updates mid-turn still has to replay from the previous turn's `turn_complete` (or fall back to `"full"`).

Stamping a `messageId` on every recordable `session/update` (`tool_call`, `tool_call_update`, `plan`, `agent_message_chunk`, …) is cheap (~24 bytes per entry) and lets clients resync at whatever granularity they last observed. State-shaped events that don't go into history (`current_model_update`, `current_mode_update`, `usage_update`, `available_commands_update`, `session_info_update`, `client_disconnected`) deliberately skip the stamp since there's no history entry for an `after_message` lookup to land on.

Worth either explicitly broadening messageId in the spec, or noting that proxies MAY stamp messageId on additional event kinds for finer resync.

### 4. Deliver attach history as live notifications instead of inline `result.history`

Lower priority — flagging as a question rather than a proposal. The RFD puts replayed history in `result.history` as an array. I chose to stream the entries as separate `session/update` notifications after the attach response so live and replayed events use the same wire path (clients have a single dispatcher). It also avoids unbounded response sizes for long sessions.

Both work. Worth a sentence in the spec acknowledging implementations MAY stream rather than inline, since the difference is observable to clients and they should know not to require one shape.

### 5. `client_connected` reciprocal to `client_disconnected`

`client_disconnected` is great for tearing down presence UI. The reciprocal — a `session/update` of `sessionUpdate: "client_connected"` carrying `{client, timestamp}` when a new peer attaches — would let presence-aware clients (dashboards, peer panels) build a live roster without polling. Today my only roster snapshot is `connectedClients` in the attach response, which goes stale the moment another client joins.

PR #541, opened 2026-02-19
RFD: Add flag config option type for boolean toggles

fscarponi:
This RFD proposes adding a new `flag` type to session configuration options, enabling agents to expose simple boolean ON/OFF toggles as first-class config options alongside the existing `select` type.

#### Motivation

Currently, the only way to expose a boolean toggle (e.g., "Brave Mode", "Read Only", "Produce Report") is to use a `select` with two artificial options like "on"/"off". This forces clients to implement custom, non-agnostic logic to detect which selects are actually boolean flags — defeating the purpose of a standardized protocol.

A dedicated `flag` type allows clients to natively render toggle switches or checkboxes without guessing.

#### Proposed changes

- `SessionConfigFlag` struct with `current_value: bool`
- `Flag` variant in `SessionConfigKind` (discriminated by `"type": "flag"`)
- `SessionConfigOptionValue` untagged enum (`String` | `Bool`) for `SetSessionConfigOptionRequest.value`
- Wire-level backward compatible: existing JSON payloads with string values remain valid

#### Working implementation

A reference implementation is available on the branch [`fabrizio.scarponi/flag-config-option`](https://github.com/fscarponi/agent-client-protocol/tree/fabrizio.scarponi/flag-config-option), including updated docs, schema regeneration, and passing tests/clippy.

Happy to open a follow-up implementation PR once this RFD is reviewed and championed


anna239:
@benbrandt what do you think? I believe this is a useful thing to have, but I don't have any opinion on naming here — whether it should be called `flag` or `boolean` or something else

docs/rfds/flag-config-option.mdx
```diff
@@ -0,0 +1,53 @@
...
+
+## Implementation details and plan
+
+A working implementation is available at: https://github.com/fscarponi/agent-client-protocol/tree/fabrizio.scarponi/flag-config-option
+
+Key changes:
+
+1. `SessionConfigFlag` struct with `current_value: bool`
+2. `Flag` variant in `SessionConfigKind` (tagged via `"type": "flag"`)
+3. `SessionConfigOptionValue` untagged enum (`String` | `Bool`) replacing `SessionConfigValueId` in `SetSessionConfigOptionRequest.value`
```
benbrandt: I'll want to look at this one closely at implementation time to make sure but I think there is an ok way to do this, especially because this is agent driven. 
An agent will only provide these types of values if they support them

docs/rfds/flag-config-option.mdx
```diff
@@ -0,0 +1,53 @@
+---
+title: "Flag Config Option Type"
+---
+
+Author(s): [fscarponi](https://github.com/fscarponi)
+
+## Elevator pitch
+
+Add a new `flag` type to session configuration options, enabling agents to expose simple boolean ON/OFF toggles (e.g., "Brave Mode", "Read Only", "Produce Report") as first-class config options alongside the existing `select` type.
```
benbrandt: Naming nitpick: I wonder if it would be better to go with checkbox since we went with select and some other html form api conventions... but checkboxes in html forms are a little weird in terms of their checked + value behavior.

Another reason to treat this is a flag instead is that html checkboxes allow for grouped values (multiple checkboxes for the same field name, basically a multi-select)

@anna239 not sure how you feel about this one. Not a strong opinion here, just asking

benbrandt: Ah sorry missed your comment :D let me do some research real quick

benbrandt: Ok looking around, I think something like `boolean`, `toggle` or `switch` or more common

Maybe we just go with boolean? thoughts?

anna239: sure, let's go with boolean, I'll update the text

PR #576, opened 2026-02-24
Add config option type for boolean on/off toggles

fscarponi:
#541 implementation

This PR adds a new `boolean` type to session configuration options, enabling agents to expose simple ON/OFF toggles (e.g., "Brave Mode", "Read Only", "Produce Report") as first-class config options alongside the existing `select` type.

All new types are gated behind the `unstable_boolean_config` feature flag, so they can be iterated on without breaking the stable API.

#### Motivation

The current `select` config option type works well for choosing between multiple values (models, modes, reasoning levels), but it doesn't naturally cover the use case of a simple on/off toggle. Without a dedicated boolean type, clients would need custom, non-agnostic logic to distinguish boolean toggles from regular selectors — defeating the purpose of a standardized protocol.

A dedicated `boolean` type allows clients to natively render toggle switches or checkboxes without guessing.

#### Changes

**Rust schema (`src/agent.rs`)** — gated behind `#[cfg(feature = "unstable_boolean_config")]`:

- Added `SessionConfigBoolean` struct with a `current_value: bool` field
- Added `Boolean(SessionConfigBoolean)` variant to `SessionConfigKind` (discriminated via `"type": "boolean"`)
- Added `boolean()` convenience constructor on `SessionConfigOption`

**Always available** (backward compatible, no-op when no boolean options exist):

- Added `SessionConfigOptionValue` enum (`#[serde(untagged)]`: `String` | `Bool`) to support both string and boolean values in `SetSessionConfigOptionRequest`
- Added `From` impls (`SessionConfigValueId`, `bool`, `&str`) for ergonomic construction
- The `SetSessionConfigOptionRequest.value` field type changed from `SessionConfigValueId` to `SessionConfigOptionValue` — source-compatible for existing callers thanks to `From` impls

**Feature flag (`Cargo.toml`)**:

- Added `unstable_boolean_config` feature, included in the `unstable` meta-feature

**Documentation (`docs/protocol/session-config-options.mdx`)**:

- Added "Boolean Options" section documenting the new type
- Added a `brave_mode` boolean example to the Initial State JSON
- Updated the `session/set_config_option` parameter docs to reflect `string | boolean` value type

**RFD (`docs/rfds/flag-config-option.mdx`)**:

- Included RFD document describing the motivation, proposal, and FAQ

**Generated schemas**:

- `schema.json` — stable, does **not** include boolean type
- `schema.unstable.json` — includes boolean type
- All derived files regenerated via `npm run generate`

#### Wire compatibility

- **Fully backward-compatible.** `SessionConfigOptionValue` uses `#[serde(untagged)]`, so existing JSON payloads with string values continue to deserialize correctly. The new boolean variant is additive and only available when the unstable feature is enabled.

#### Verification

- `cargo clippy` (no features) — no warnings
- `cargo clippy --all-features` — no warnings
- `cargo test` — 26 tests + 2 doctests pass (stable)
- `cargo test --all-features` — 27 tests + 2 doctests pass (with boolean)
- `npm run generate` — schemas regenerated successfully
- `prettier --check .` / `cargo fmt -- --check` — formatting clean

### DIFF
--- a/docs/protocol/draft/schema.mdx
+++ b/docs/protocol/draft/schema.mdx
@@ -726,27 +726,41 @@ Sets the current value for a session configuration option.
 
 Request parameters for setting a session configuration option.
 
-**Type:** Object
+**Type:** Union
 
-**Properties:**
+<ResponseField name="boolean" type="object">
+A boolean value (`type: "boolean"`).
 
-<ResponseField name="_meta" type={"object | null"} >
-  The _meta property is reserved by ACP to allow clients and agents to attach additional
-metadata to their interactions. Implementations MUST NOT make assumptions about values at
-these keys.
+<Expandable title="Properties">
 
-See protocol docs: [Extensibility](https://agentclientprotocol.com/protocol/extensibility)
-
+<ResponseField name="type" type={"string"} required>
+  The discriminator value. Must be `"boolean"`.
 </ResponseField>
-<ResponseField name="configId" type={<a href="#sessionconfigid">SessionConfigId</a>} required>
-  The ID of the configuration option to set.
+<ResponseField name="value" type={"boolean"} required>
+  The boolean value.
 </ResponseField>
-<ResponseField name="sessionId" type={<a href="#sessionid">SessionId</a>} required>
-  The ID of the session to set the configuration option for.
+
+</Expandable>
 </ResponseField>
-<ResponseField name="value" type={<a href="#sessionconfigoptionvalue">SessionConfigOptionValue</a>} required>
-  The value to set. For `select` options, this is a `SessionConfigValueId` string.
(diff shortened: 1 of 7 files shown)


benbrandt:
@anna239 @fscarponi want to take a look at this version? 
I added an optional type field so that in the future we can distinguish between normal string values and ids

In hindsight we should have had this value type all along :D but we have learned, I will definitely take a close look for v2 to include these sorts of things

anna239:
lgtm, @fscarponi could you please update kotlin-sdk implementation accordingly?

fscarponi:
Ok fine! i'll update the kotlin sdk soon! Thanks @benbrandt

PR #582, opened 2026-02-25
rfd: add dynamic MCP server updates in live sessions

OmChillure:
Description:

This PR introduces an RFD for adding a new session/update_servers request to the Agent Client Protocol.

Motivation: Currently, MCP servers can only be provided during session setup (session/new, session/load, etc.). Once a session is established, its list of MCP servers is fixed. This forces users to restart their sessions (losing valuable context and history) just to add, remove, or reconfigure an MCP server.

Proposed Changes:

Introduce a session/update_servers request sent by the Client to the Agent.
The request payload contains the full replacement list of MCP servers (reusing the existing 

McpServer
 schemas).
The Agent gracefully connects to newly introduced servers and disconnects from those removed from the list.
Impact: This allows clients to confidently provide live MCP server management UIs. Settings changes to MCP configurations can instantly propagate to active agent sessions without interrupting the user's workflow.

Based on the feature request in discussion [#556](https://github.com/orgs/agentclientprotocol/discussions/556).

### DIFF
--- a/docs/rfds/mcp-server-live-update.mdx
+++ b/docs/rfds/mcp-server-live-update.mdx
@@ -1,0 +1,176 @@
+---
+title: "Dynamic MCP Server Updates in Live Sessions"
+---
+
+Author(s): [OmChillure](https://github.com/OmChillure)
+
+> This RFD is based on a feature request originally raised by [@dlqqq](https://github.com/dlqqq) in [discussion #556](https://github.com/orgs/agentclientprotocol/discussions/556).
+
+## Elevator pitch
+
+We propose adding a new `session/update_servers` request that allows ACP Clients to send a replacement list of MCP servers to an Agent during an active session — without requiring a session restart. This unblocks a class of client UX where users can add, remove, or reconfigure MCP servers mid-conversation and have the change take effect immediately.
+
+## Status quo
+
+Currently, ACP Agents are only informed of available MCP servers during session setup — in the `session/new`, `session/load`, and `session/resume` requests. Once a session is established, the MCP server list is fixed for its entire lifetime. There is no mechanism to update it afterwards.
+
+This creates several problems:
+
+* **Session restarts required**: Adding, removing, or reconfiguring an MCP server mid-session requires the user to end their current session and start a new one — losing all conversation context and state.
+* **Client UIs are hobbled**: Clients with live MCP server management panels (toggle a server on/off, install an extension) cannot make those changes take effect in the current session. The user must restart to see any effect.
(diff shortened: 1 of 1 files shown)


docs/rfds/mcp-server-live-update.mdx
```diff
@@ -0,0 +1,176 @@
...
+
+* **Session restarts required**: Adding, removing, or reconfiguring an MCP server mid-session requires the user to end their current session and start a new one — losing all conversation context and state.
+* **Client UIs are hobbled**: Clients with live MCP server management panels (toggle a server on/off, install an extension) cannot make those changes take effect in the current session. The user must restart to see any effect.
+* **Workflow interruption**: A user who realizes mid-task that they need an additional tool (e.g. a GitHub MCP server when they need to open a PR) must abandon the conversation.
+* **Inconsistency**: Other session properties — modes, models, config options — can already be updated dynamically in a live session. MCP server availability is the only major session property that cannot.
+
+## What we propose to do about it
+
+We propose adding a new `session/update_servers` request type, sent by the ACP Client to the Agent during an active session. The request carries the full updated list of MCP servers, which replaces the prior list entirely. The Agent connects to any newly added servers and disconnects from any that are no longer present.
```
benbrandt, line 26: @ignatov I think in some ways this is tied to your intiative to clarify MCP lifecycle in general. i.e should it replace all? 
Should we instead have a add/remove endpoint?
And a way to retrieve all currently configured MCPs in case there were some from other agent config?

docs/rfds/mcp-server-live-update.mdx
```diff
@@ -0,0 +1,176 @@
...
+title: "Dynamic MCP Server Updates in Live Sessions"
+---
+
+Author(s): [OmChillure](https://github.com/OmChillure)
+
+> This RFD is based on a feature request originally raised by [@dlqqq](https://github.com/dlqqq) in [discussion #556](https://github.com/orgs/agentclientprotocol/discussions/556).
+
+## Elevator pitch
+
+We propose adding a new `session/update_servers` request that allows ACP Clients to send a replacement list of MCP servers to an Agent during an active session — without requiring a session restart. This unblocks a class of client UX where users can add, remove, or reconfigure MCP servers mid-conversation and have the change take effect immediately.
```
benbrandt, line 11: nitpick, I feel like this should be session/update_mcp_servers to be clearer what we are updating

OmChillure: Yeah would make better sense

PR #588, opened 2026-02-27
feat(unstable): Initial support for various auth methods

benbrandt:
In building this out, and looking at some implementations, it became clear we should have some of these be opt-in by the client so that it isn't showing auth methods it can't feasibly support.

### DIFF
--- a/docs/rfds/auth-methods.mdx
+++ b/docs/rfds/auth-methods.mdx
@@ -2,7 +2,7 @@
 title: "Authentication Methods"
 ---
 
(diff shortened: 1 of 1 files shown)


benbrandt:
@anna239 I added a client capability section for auth method types so that the agent only sends ones the client can support.

This will also be helpful if we add future ones that require additional features.

I didn't require a capability for the env var because we are mostly providing extra information for that case to present to the user. The client ideally would have a way to send that env var to the agent, but I think it is ok.

Thoughts?

anna239:
@benbrandt looks great to me, thank you!

PR #605, opened 2026-03-01
fix: allow null value for used field in UsageUpdate

sternelee:
The `used` field in UsageUpdate can be null when the agent doesn't provide token usage information. Changed from `u64` to `Option<u64>` to handle this case gracefully.

Fixes decode error: "invalid type: null, expected u64"


benbrandt:
What agent is this? If they don't support usage why are they sending this at all?

Issue #554, opened 2026-02-20
Add turn-complete signal for session_update notifications


simonrosenberg:
## Problem

After a `prompt()` call returns, there is no protocol-level signal indicating that all `session_update` notifications for that turn have been delivered. Clients that accumulate state from these notifications (e.g. streamed text chunks, usage updates) must resort to heuristic delays (`asyncio.sleep`) to wait for in-flight notifications to arrive before reading the accumulated result.

This is inherently racy — too short a delay misses late notifications, too long a delay adds unnecessary latency.

## Current workaround

In [OpenHands](https://github.com/OpenHands/software-agent-sdk), we yield to the event loop and then sleep for a configurable duration (default 0.1s) after `prompt()` returns:

```python
response = await conn.prompt(messages, session_id)
await asyncio.sleep(0)  # yield to let queued handlers run
await asyncio.sleep(0.1)  # wait for in-flight notifications
```

## Proposed solution

Add a protocol-level mechanism that signals when all notifications for a given turn have been delivered. For example:

- A dedicated `turn_complete` notification sent by the server after all `session_update` notifications for a prompt response
- Or a field on the `PromptResponse` indicating whether all associated notifications have been flushed

This would let clients reliably synchronize without arbitrary delays.

benbrandt:
Hi 👋🏻 yes I think I would like to move the entire prompt lifecycle to be notification based entirely, which I think will help with this.

denofevil:
Relates to https://github.com/agentclientprotocol/agent-client-protocol/pull/533/changes/7375612a51e87b08a36eb4667ec25eece26dd59e#diff-d916e83e9af6279a58026089bd8758f3a0884148c3728e1b475106cc33f68df2R215

stablegenius49:
I opened a draft RFD PR for this: https://github.com/agentclientprotocol/agent-client-protocol/pull/644\n\nIt proposes a   barrier (with capability gating) as an RFD-first step before unstable schema/runtime changes.

stablegenius49:
Follow-up clarification: the proposal is a TURN_COMPLETE session update barrier (plus sessionCapabilities.turnComplete capability advertisement) as an RFD-first step before unstable schema/runtime changes.

xenodium:
> Add a protocol-level mechanism that signals when all notifications for a given turn have been delivered. For example:
>  A dedicated turn_complete notification sent by the server after all session_update notifications for a prompt response

[...]

PR #644, opened 2026-03-04
docs(rfd): draft turn_complete signal for session/update sync

stablegenius49:
## Summary

Draft RFD to address #554 by proposing a protocol-level `turn_complete` signal in `session/update`.

This gives clients a deterministic end-of-turn barrier for streamed updates and removes the need for sleep-based heuristics after `session/prompt` resolves.

## What this PR changes

- Adds a new Draft RFD: `docs/rfds/turn-complete-signal.mdx`
- Adds the RFD to docs navigation (`docs/docs.json` under **RFDs > Draft**)

## Proposal highlights

- New update variant: `sessionUpdate: "turn_complete"`
- Emitted once per prompt turn, after all other updates for that turn
- Includes turn correlation (`promptRequestId`) and optional `stopReason`
- Explicit capability advertisement via `sessionCapabilities.turnComplete`

## Why RFD-first

This appears to touch ordering semantics and SDK behavior across multiple implementations. Starting with an RFD should help converge on semantics before unstable schema/runtime changes.

Refs #554

### DIFF
--- a/docs/docs.json
+++ b/docs/docs.json
@@ -123,7 +123,8 @@
               "rfds/diff-delete",
               "rfds/boolean-config-option",
               "rfds/elicitation",
-              "rfds/session-stop"
+              "rfds/session-stop",
+              "rfds/turn-complete-signal"
             ]
           },
           {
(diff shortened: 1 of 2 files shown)


benbrandt:
Thanks for looking into this! Given where we are going in v2 with the new prompt lifecycle though https://agentclientprotocol.com/rfds/v2/prompt, I don't think we'll add this.

PR #648, opened 2026-03-04
docs(rfd): Сustom LLM endpoints

xtmq:
---
title: "Configurable LLM Providers"
---

- Author(s): [@anna239](https://github.com/anna239), [@xtmq](https://github.com/xtmq)

## Elevator pitch

Add the ability for clients to discover and configure agent LLM providers (identified by `id`) via dedicated provider methods:

- `providers/list`
- `providers/set`
- `providers/disable`

This allows clients to route LLM requests through their own infrastructure (proxies, gateways, or self-hosted models) without agents needing to know about this configuration in advance.

## Status quo

ACP does not currently define a standard method for configuring LLM providers.

In practice, provider configuration is usually done via environment variables or agent-specific config files. That creates several problems:

- No standard way for clients to discover what providers an agent exposes
- No standard way to update one specific provider by id
- No standard way to disable a specific provider at runtime while preserving provider discoverability
- Secret-bearing values in headers are difficult to handle safely when configuration must be round-tripped

This particularly affects:

- **Client proxies**: clients want to route agent traffic through their own proxies, for example to add headers or logging
- **Enterprise deployments**: organizations want to route LLM traffic through internal gateways for compliance, logging, and cost controls
- **Self-hosted models**: users running local servers (vLLM, Ollama, etc.) need to redirect agent traffic to local infrastructure
- **API gateways**: organizations using multi-provider routing, rate limiting, and caching need standardized endpoint configuration

## Shiny future

Clients will be able to:

1. Understand whether an agent supports client-managed LLM routing
2. See where the agent is currently sending LLM requests (for example in settings UI)
3. Route agent LLM traffic through their own infrastructure (enterprise proxy, gateway, self-hosted stack)
4. Update routing settings from the client instead of relying on agent-specific env vars
5. Disable a provider when needed and later re-enable it explicitly
6. Apply these settings before starting new work in sessions

### Intended flow

```mermaid
sequenceDiagram
    participant Client
    participant Agent

    Client->>Agent: initialize
    Agent-->>Client: initialize response (agentCapabilities.providers = true)

    Client->>Agent: providers/list
    Agent-->>Client: providers/list response

    Client->>Agent: providers/set (id = "main")
    Agent-->>Client: providers/set response

    Client->>Agent: providers/disable (optional)
    Agent-->>Client: providers/disable response

    Client->>Agent: session/new
```

1. Client initializes and checks `agentCapabilities.providers`.
2. Client calls `providers/list` to discover available providers, their current routing targets (or disabled state), supported protocol types, and whether they are required.
3. Client calls `providers/set` to apply new (required) configuration for a specific provider id.
4. Client may call `providers/disable` when a non-required provider should be disabled.
5. Client creates or loads sessions.

### Capability advertisement

The agent advertises support with a single boolean capability:

```typescript
interface AgentCapabilities {
  // ... existing fields ...

  /**
   * Provider configuration support.
   * If true, the agent supports providers/list, providers/set, and providers/disable.
   */
  providers?: boolean;
}
```

If `providers` is absent or `false`, clients must treat provider methods as unsupported.

### Types

```typescript
/** Well-known API protocol identifiers. */
type LlmProtocol = "anthropic" | "openai" | "azure" | "vertex" | "bedrock";

interface ProviderCurrentConfig {
  /** Protocol currently used by this provider. */
  apiType: LlmProtocol;

  /** Base URL currently used by this provider. */
  baseUrl: string;
}

interface ProviderInfo {
  /** Provider identifier, for example "main" or "openai". */
  id: string;

  /** Supported protocol types for this provider. */
  supported: LlmProtocol[];

  /**
   * Whether this provider is mandatory and cannot be disabled via providers/disable.
   * If true, clients must not call providers/disable for this id.
   */
  required: boolean;

  /**
   * Current effective non-secret routing config.
   * Null means provider is disabled.
   */
  current: ProviderCurrentConfig | null;

  /** Extension metadata */
}
```

### `providers/list`

```typescript
interface ProvidersListRequest {
  /** Extension metadata */
}

interface ProvidersListResponse {
  /** Configurable providers with current routing info suitable for UI display. */
  providers: ProviderInfo[];

  /** Extension metadata */
}
```

### `providers/set`

`providers/set` updates the full configuration for one provider id.

```typescript
interface ProvidersSetRequest {
  /** Provider id to configure. */
  id: string;

  /** Protocol type for this provider. */
  apiType: LlmProtocol;

  /** Base URL for requests sent through this provider. */
  baseUrl: string;

  /**
   * Full headers map for this provider.
   * May include authorization, routing, or other integration-specific headers.
   */
  headers: Record<string, string>;

  /** Extension metadata */
}

interface ProvidersSetResponse {
  /** Extension metadata */
}
```

### `providers/disable`

```typescript
interface ProvidersDisableRequest {
  /** Provider id to disable. */
  id: string;

  /** Extension metadata */
}

interface ProvidersDisableResponse {
  /** Extension metadata */
}
```

### Example exchange

**initialize Response:**

```json
{
  "jsonrpc": "2.0",
  "id": 0,
  "result": {
    "protocolVersion": 1,
    "agentInfo": {
      "name": "MyAgent",
      "version": "2.0.0"
    },
    "agentCapabilities": {
      "providers": true,
      "sessionCapabilities": {}
    }
  }
}
```

**providers/list Request:**

```json
{
  "jsonrpc": "2.0",
  "id": 1,
  "method": "providers/list",
  "params": {}
}
```

**providers/list Response:**

```json
{
  "jsonrpc": "2.0",
  "id": 1,
  "result": {
    "providers": [
      {
        "id": "main",
        "supported": ["bedrock", "vertex", "azure", "anthropic"],
        "required": true,
        "current": {
          "apiType": "anthropic",
          "baseUrl": "http://localhost/anthropic"
        }
      },
      {
        "id": "openai",
        "supported": ["openai"],
        "required": false,
        "current": null
      }
    ]
  }
}
```

**providers/set Request:**

```json
{
  "jsonrpc": "2.0",
  "id": 2,
  "method": "providers/set",
  "params": {
    "id": "main",
    "apiType": "anthropic",
    "baseUrl": "https://llm-gateway.corp.example.com/anthropic/v1",
    "headers": {
      "X-Request-Source": "my-ide"
    }
  }
}
```

**providers/set Response:**

```json
{
  "jsonrpc": "2.0",
  "id": 2,
  "result": {}
}
```

**providers/disable Request:**

```json
{
  "jsonrpc": "2.0",
  "id": 3,
  "method": "providers/disable",
  "params": {
    "id": "openai"
  }
}
```

**providers/disable Response:**

```json
{
  "jsonrpc": "2.0",
  "id": 3,
  "result": {}
}
```

### Behavior

1. **Capability discovery**: agents that support provider methods MUST advertise `agentCapabilities.providers: true` in `initialize`. Clients SHOULD only call `providers/*` when this capability is present and `true`.
2. **Timing and session impact**: provider methods MUST be called after `initialize`. Clients SHOULD configure providers before creating or loading sessions. Agents MAY choose not to apply changes to already running sessions, but SHOULD apply them to sessions created or loaded after the change.
3. **List semantics**: `providers/list` returns configurable providers, their supported protocol types, current effective routing, and `required` flag. Providers SHOULD remain discoverable in list after `providers/disable`.
4. **Client behavior for required providers**: clients SHOULD NOT call `providers/disable` for providers where `required: true`.
5. **Disabled state encoding**: in `providers/list`, `current: null` means the provider is disabled and MUST NOT be used by the agent for LLM calls.
6. **Set semantics and validation**: `providers/set` replaces the full configuration for the target `id` (`apiType`, `baseUrl`, full `headers`). If `id` is unknown, `apiType` is unsupported for that provider, or params are malformed, agents SHOULD return `invalid_params`.
7. **Disable semantics**: `providers/disable` disables the target provider at runtime. A disabled provider MUST appear in `providers/list` with `current: null`. If target provider has `required: true`, agents MUST return `invalid_params`. Disabling an unknown `id` SHOULD be treated as success (idempotent behavior).
8. **Scope and persistence**: provider configuration is process-scoped and SHOULD NOT be persisted to disk.

## Frequently asked questions

### What does `null` mean in `providers/list`?

`current: null` means the provider is disabled.

When disabled, the agent MUST NOT route LLM calls through that provider until the client enables it again with `providers/set`.

### Why is there a `required` flag?

Some providers are mandatory for agent operation and must not be disabled.

`required` lets clients hide or disable the provider-disable action in UI and avoid calling `providers/disable` for those ids.

### Why not a single `providers/update` method for full list replacement?

A full-list update means the client must send complete configuration (including `headers`) for all providers every time.

If the client wants to change only one provider, it may not know headers for the others. In that case it cannot safely build a correct full-list payload.

Also, `providers/list` does not return headers, so the client cannot simply "take what the agent returned" and send it back with one edit.

Per-provider methods (`set` and `disable`) avoid this problem and keep updates explicit.

### Why doesn't `providers/list` return headers?

Header values may contain secrets and should not be echoed by the agent. `providers/list` is intentionally limited to non-secret routing information (`current.apiType`, `current.baseUrl`).

### Why are `providers/list` and `providers/set` payloads different?

`providers/set` accepts `headers`, including secrets, and is write-oriented.

`providers/list` is read-oriented and returns only non-secret routing summary (`current`) for UI and capability discovery.

### Why is this separate from `initialize` params?

Clients need capability discovery first, then provider discovery, then configuration. A dedicated method family keeps initialization focused on negotiation and leaves provider mutation to explicit steps.

### Why not use `session-config` with a `provider` category instead?

`session-config` is a possible alternative, and we may revisit it as the spec evolves.

We did not choose it as the primary approach in this proposal because provider routing here needs dedicated semantics that are difficult to express with today's session config model:

- Multiple providers identified by `id`, each with its own lifecycle
- Structured payloads (`apiType`, `baseUrl`, full `headers` map) rather than simple scalar values
- Explicit discoverable (`providers/list`) and disable (`providers/disable`) semantics

Today, `session-config` values are effectively string-oriented and do not define a standard multi-value/structured model for this use case.

## Revision history

- 2026-03-22: Finalized provider disable semantics - `providers/remove` renamed to `providers/disable`, required providers are non-disableable, and disabled state is represented as `current: null`
- 2026-03-21: Initial draft of provider configuration API (`providers/list`, `providers/set`, `providers/remove`)
- 2026-03-07: Rename "provider" to "protocol" to reflect API compatibility level; make `LlmProtocol` an open string type with well-known values; resolve open questions on identifier standardization and model availability
- 2026-03-04: Revised to use dedicated `setLlmEndpoints` method with capability advertisement
- 2026-02-02: Initial draft - preliminary proposal to start discussion

### DIFF
--- a/docs/rfds/custom-llm-endpoint.mdx
+++ b/docs/rfds/custom-llm-endpoint.mdx
@@ -1,0 +1,291 @@
+---
+title: "Custom LLM Endpoint Configuration"
+---
+
+- Author(s): [@anna239](https://github.com/anna239)
+
+> **Note:** This RFD is very preliminary and intended to start a dialog about this feature. The proposed design may change significantly based on feedback and further discussion.
+
+## Elevator pitch
+
+> What are you proposing to change?
+
+Add the ability for clients to pass a custom LLM endpoint URL and authentication credentials to agents during initialization. This allows organizations to route all LLM requests through their own infrastructure (proxies, gateways, or self-hosted models) without agents needing to know about this configuration in advance.
+
+## Status quo
+
+> How do things work today and what problems does this cause? Why would we change things?
+
+Currently, agents are configured with their own LLM endpoints and credentials, typically through environment variables or configuration files. This creates problems for:
+
+- **Client proxies**: Clients want to route agent traffic through their own proxies, f.i. for setting additional headers or logging
+- **Enterprise deployments**: Organizations want to route LLM traffic through their own proxies for compliance, logging, or cost management
+- **Self-hosted models**: Users running local LLM servers (vLLM, Ollama, etc.) cannot easily redirect agent traffic
+- **API gateways**: Organizations using LLM gateways for rate limiting, caching, or multi-provider routing
+
+## Shiny future
+
+> How will things play out once this feature exists?
+
+Clients will be able to:
+1. Pass a custom LLM endpoint URL and authentication header
+2. Have all agent LLM requests automatically routed through that endpoint
+
+## Implementation details and plan
+
+> Tell me more about your implementation. What is your detailed implementation plan?
+
+We present three alternative approaches for discussion.
+
+### Alternative A: Pass in `initialize` request
+
+Add a new optional `llmEndpoint` property to `InitializeRequest`:
+
+```typescript
+interface InitializeRequest {
+  // ... existing fields ...
+
+  /**
+   * Custom LLM endpoint configuration.
+   * When provided, the agent should route all LLM requests through this endpoint.
+   * This configuration is per-process and should not be persisted to disk.
+   */
+  llmEndpoint?: LlmEndpointConfig | null;
+}
+
+interface LlmEndpointConfig {
+  /** Base URL for LLM API requests (e.g., "https://llm-proxy.corp.example.com/v1") */
+  url: string;
+
+  /**
+   * Value for the Authorization header.
+   * Will be sent as-is (e.g., "Bearer sk-..." or "Basic ...").
+   */
+  authHeader: string;
+
+  /** Extension metadata */
+  _meta?: Record<string, unknown>;
+}
+
+interface InitializeResponse {
+  // ... existing fields ...
+
+  /**
+   * Echoed back when the agent accepts the custom LLM endpoint.
+   * If present in the response, the agent confirms it will use this endpoint.
+   * If absent (and was provided in request), the agent did not accept it.
+   */
+  llmEndpoint?: LlmEndpointConfig | null;
+}
+```
+
+#### JSON Schema Additions
+
+```json
+{
+  "$defs": {
+    "LlmEndpointConfig": {
+      "description": "Configuration for a custom LLM endpoint. This configuration is per-process and should not be persisted to disk.",
+      "properties": {
+        "url": {
+          "type": "string",
+          "description": "Base URL for LLM API requests."
+        },
+        "authHeader": {
+          "type": ["string"],
+          "description": "Value for the Authorization header. Sent as-is."
+        },
+        "_meta": {
+          "additionalProperties": true,
+          "type": ["object", "null"]
+        }
+      },
+      "required": ["url", "authHeader"],
+      "type": "object"
+    }
+  }
+}
+```
+
+#### Example Exchange
+
+**Initialize Request:**
+```json
+{
+  "jsonrpc": "2.0",
+  "id": 1,
+  "method": "initialize",
+  "params": {
+    "clientInfo": {
+      "name": "MyIDE",
+      "version": "1.0.0"
+    },
+    "protocolVersion": "2025-01-01",
+    "llmEndpoint": {
+      "url": "https://llm-gateway.corp.example.com/v1",
+      "authHeader": "Bearer temporary-token-abc123"
+    }
+  }
+}
+```
+
+**Initialize Response (endpoint accepted):**
+```json
+{
+  "jsonrpc": "2.0",
+  "id": 1,
+  "result": {
+    "protocolVersion": "2025-01-01",
+    "agentInfo": {
+      "name": "MyAgent",
+      "version": "2.0.0"
+    },
+    "agentCapabilities": {
+      "sessionCapabilities": {}
+    },
+    "llmEndpoint": {
+      "url": "https://llm-gateway.corp.example.com/v1",
+      "authHeader": "Bearer temporary-token-abc123"
+    }
+  }
+}
+```
+
+**Initialize Response (endpoint not accepted):**
+```json
+{
+  "jsonrpc": "2.0",
+  "id": 1,
+  "result": {
+    "protocolVersion": "2025-01-01",
+    "agentInfo": {
+      "name": "MyAgent",
+      "version": "2.0.0"
+    },
+    "agentCapabilities": {
+      "sessionCapabilities": {}
+    }
+  }
+}
+```
+
+#### Behavior
+
+1. **Confirmation via response**: When the agent accepts the `llmEndpoint`, it MUST echo it back in the `InitializeResponse`. If `llmEndpoint` is absent from the response, the client should assume the agent did not accept the custom endpoint configuration.
+
+2. **Per-process scope**: The `llmEndpoint` configuration applies to the entire agent process lifetime. It should not be stored to disk or persist beyond the process.
+
+3. **All LLM requests**: When `llmEndpoint` is accepted, the agent should route ALL LLM API requests through the specified URL, using the provided `authHeader` for authentication.
+
+4. **No mid-execution changes**: The endpoint cannot be changed after initialization. If a different endpoint is needed, a new agent process must be started.
+
+5. **Agent discretion**: If an agent cannot support custom endpoints (e.g., uses a proprietary API), it should omit `llmEndpoint` from the response or return an error during initialization.
+
+### Alternative B: Dedicated `setLlmEndpoint` method
+
+Instead of passing the endpoint in `initialize`, introduce a dedicated method that can be called after initialization but before session creation.
+
+```typescript
+interface SetLlmEndpointRequest {
+  /** Base URL for LLM API requests */
+  url: string;
+
+  /** Value for the Authorization header */
+  authHeader: string;
+
+  /** Extension metadata */
+  _meta?: Record<string, unknown>;
+}
+
+interface SetLlmEndpointResponse {
+  /** Whether the endpoint was accepted */
+  accepted: boolean;
+
+  /** Extension metadata */
+  _meta?: Record<string, unknown>;
+}
+```
+
+**Example flow:**
+1. Client calls `initialize`
+2. Client calls `setLlmEndpoint` with URL and auth header
+3. Agent confirms acceptance
+4. Client creates session
+
+**Trade-offs:**
+- (+) Cleaner separation of concerns - initialization vs configuration
+- (+) Could potentially allow changing endpoint between sessions (if desired)
+- (-) Different sessions might want different endpoints, but we want a single endpoint per process
+
+### Alternative C: Registry-declared environment variables
+
+Declare in the Agent Registry which environment variables or command-line arguments should be used to pass the URL and token when starting the agent process.
+
+```json
+{
+  "name": "MyAgent",
+  "command": "my-agent",
+  "llmEndpointConfig": {
+    "urlEnvVar": "LLM_BASE_URL",
+    "authEnvVar": "LLM_AUTH_TOKEN"
+  }
+}
+```
+
+Or with command-line arguments:
+
+```json
+{
+  "name": "MyAgent",
+  "command": "my-agent",
+  "llmEndpointConfig": {
+    "urlArg": "--llm-url",
+    "authArg": "--llm-token"
+  }
+}
+```
+
+The client would then start the agent process with:
+```bash
+LLM_BASE_URL="https://gateway.example.com/v1" LLM_AUTH_TOKEN="Bearer sk-..." my-agent
+# or
+my-agent --llm-url "https://gateway.example.com/v1" --llm-token "Bearer sk-..."
+```
+
+**Trade-offs:**
+- (+) Agents in the Registry can be marked as supporting custom endpoints
+- (+) Easier to implement for agents that already read these values from environment variables
+- (+) No protocol changes needed - uses existing process spawning mechanisms
+- (+) Works even before ACP connection is established
+- (-) Requires Registry support
+- (-) Less dynamic - configuration is fixed at process start
+
+## Open questions
+
+### How should model availability be handled?
+
+When a custom endpoint is provided, it may only support a subset of models. For example, a self-hosted vLLM server might only have `llama-3-70b` available, while the agent normally advertises `claude-3-opus`, `gpt-4`, etc.
+
+### Should the agent advertise support for custom endpoints?
+
+Should there be a capability flag indicating the agent supports `llmEndpoint`? This would let clients know whether passing this configuration will have any effect.
+
+## Frequently asked questions
+
+> What questions have arisen over the course of authoring this document?
+
+### Why not pass endpoint when selecting a model?
+
+One option would be to pass the endpoint URL and credentials when the user selects a model (e.g., in `session/new` or a model selection method).
+
+Many agents throw authentication errors before the model selection happens. This makes the flow unreliable.
+
+### What if the agent doesn't support custom endpoints?
+
+If the agent doesn't support custom endpoints, it should omit `llmEndpoint` from the `InitializeResponse`. The client can then detect this by comparing the request and response:
+- If `llmEndpoint` was sent but not returned, the agent did not accept it
+- The client can then decide whether to proceed (using agent's default endpoint) or abort
+
+## Revision history
+
+- 2026-02-02: Initial draft - preliminary proposal to start discussion
--- a/docs/rfds/custom-llm-endpoint.mdx
+++ b/docs/rfds/custom-llm-endpoint.mdx
@@ -61,7 +61,7 @@ interface LlmEndpointConfig {
    * Value for the Authorization header.
    * Will be sent as-is (e.g., "Bearer sk-..." or "Basic ...").
    */
-  authHeader: string;
+  authHeader?: string | null;
 
   /** Extension metadata */
   _meta?: Record<string, unknown>;
@@ -92,7 +92,7 @@ interface InitializeResponse {
           "description": "Base URL for LLM API requests."
         },
         "authHeader": {
-          "type": ["string"],
+          "type": ["string", "null"],
           "description": "Value for the Authorization header. Sent as-is."
         },
         "_meta": {
@@ -100,7 +100,7 @@ interface InitializeResponse {
           "type": ["object", "null"]
         }
       },
-      "required": ["url", "authHeader"],
+      "required": ["url"],
       "type": "object"
     }
   }
@@ -175,7 +175,7 @@ interface InitializeResponse {
 
 2. **Per-process scope**: The `llmEndpoint` configuration applies to the entire agent process lifetime. It should not be stored to disk or persist beyond the process.
 
-3. **All LLM requests**: When `llmEndpoint` is accepted, the agent should route ALL LLM API requests through the specified URL, using the provided `authHeader` for authentication.
+3. **All LLM requests**: When `llmEndpoint` is accepted, the agent should route ALL LLM API requests through the specified URL, using the provided `authHeader` for authentication (if provided).
 
 4. **No mid-execution changes**: The endpoint cannot be changed after initialization. If a different endpoint is needed, a new agent process must be started.
 
@@ -191,7 +191,7 @@ interface SetLlmEndpointRequest {
   url: string;
 
   /** Value for the Authorization header */
-  authHeader: string;
+  authHeader?: string | null;
 
   /** Extension metadata */
   _meta?: Record<string, unknown>;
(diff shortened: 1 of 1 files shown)


docs/rfds/custom-llm-endpoint.mdx
```diff
@@ -0,0 +1,293 @@
...
+### Capability advertisement
+
+The agent advertises support for custom LLM endpoints via a new capability flag in `agentCapabilities`:
+
+```typescript
+interface AgentCapabilities {
+  // ... existing fields ...
+
+  /** Whether the agent supports custom LLM endpoint configuration via setLlmEndpoints */
+  llmEndpoints?: boolean;
```
anna239: can we also add supported providers here? like 
```
"llmEndpoints": {
    "providers": {
        "openai": {},
        "anthropic": {}
         ....
     }
}
```

xtmq: Fixed!

docs/rfds/custom-llm-endpoint.mdx
```diff
@@ -0,0 +1,292 @@
...
+}
+
+interface SetLlmEndpointsRequest {
+  /**
+   * Custom LLM endpoint configurations per provider.
+   * When provided, the agent should route LLM requests to the appropriate endpoint
+   * based on the provider being used.
+   * This configuration is per-process and should not be persisted to disk.
+   */
+  endpoints: Record<LlmProvider, LlmEndpointConfig>;
```
benbrandt: Cool, so the agent can provide multiple possible endpoints, and the client can set all of them
So for agents supporting models from multiple providers, they can set them all at once?

docs/rfds/custom-llm-endpoint.mdx
```diff
@@ -0,0 +1,292 @@
...
+3. **Endpoint configuration**: The client calls `setLlmEndpoints` with endpoint configurations for the supported providers.
+4. **Session creation**: The client proceeds to create a session.
+
+### Capability advertisement
+
+The agent advertises support for custom LLM endpoints and lists its supported providers via a new `llmEndpoints` capability in `agentCapabilities`:
+
+```typescript
+/** Well-known LLM provider identifiers */
+type LlmProvider = "anthropic" | "openai" | "google" | "amazon" | "will be added later";
```
benbrandt: Is the goal here to have "api compatibility"?
i.e. I see anthropic and therefore know that I can provide any anthropic compatible endpoint?

My question being: this could become a huge list, and I wonder if this should be driven more by endpoint compatibility?
though I guess depending on the provider you might need more or less auth requirements?

docs/rfds/custom-llm-endpoint.mdx
```diff
@@ -0,0 +1,292 @@
...
+  "id": 2,
+  "result": {}
+}
+```
+
+#### Behavior
+
+1. **Capability discovery**: The agent MUST list its supported providers in `agentCapabilities.llmEndpoints.providers` if it supports the `setLlmEndpoints` method. Clients SHOULD only send endpoint configurations for providers listed there.
+
+2. **Timing**: The `setLlmEndpoints` method MUST be called after `initialize` and before `session/new`. Calling it during an active session is undefined behavior. All subsequent sessions will use the newly configured endpoints.
```
benbrandt: I wonder if we can say: 
- Calling this MAY NOT affect currently sessions
- Agents MUST apply these settings to any sessions created or loaded after this has been called

docs/rfds/custom-llm-endpoint.mdx
```diff
@@ -0,0 +1,292 @@
...
+
+3. **Per-process scope**: The endpoint configuration applies to the entire agent process lifetime. It should not be stored to disk or persist beyond the process.
```
benbrandt: i.e it is up to the client to pass this again the next time a connection is made?
I like that, just clarifying

docs/rfds/custom-llm-endpoint.mdx
```diff
@@ -0,0 +1,292 @@
...
+
+4. **Provider-based routing**: The agent should route LLM requests to the appropriate endpoint based on the provider. If the agent uses a provider not in the provided map, it uses its default endpoint for that provider.
```
benbrandt: Should an agent ALWAYS have a default endpoint? (I think so, for simplicity + backwards compat)

docs/rfds/custom-llm-endpoint.mdx
```diff
@@ -0,0 +1,292 @@
...
+### How should provider identifiers be standardized?
+
+We need to define a standard set of provider identifiers (e.g., `"anthropic"`, `"openai"`, `"google"`, `"amazon"`). Should this be:
+- A fixed enum in the protocol specification?
+- An extensible set with well-known values and support for custom strings?
+- Defined in a separate registry/document that can be updated independently?
+
+### How should model availability be handled?
+
+When a custom endpoint is provided, it may only support a subset of models. For example, a self-hosted vLLM server might only have `llama-3-70b` available, while the agent normally advertises `claude-3-opus`, `gpt-4`, etc.
```
benbrandt: I would think many agents would be able to generate a model list based on the new provider in most cases
For those that don't they will likely show an error if you configured a provider without model support

docs/rfds/custom-llm-endpoint.mdx
```diff
@@ -0,0 +1,292 @@
...
+
+4. **Provider-based routing**: The agent should route LLM requests to the appropriate endpoint based on the provider. If the agent uses a provider not in the provided map, it uses its default endpoint for that provider.
+
+5. **Agent discretion**: If an agent cannot support custom endpoints (e.g., uses a proprietary API), it should omit `llmEndpoints` from capabilities or return an empty providers map.
+
+## Open questions
+
+### How should provider identifiers be standardized?
+
+We need to define a standard set of provider identifiers (e.g., `"anthropic"`, `"openai"`, `"google"`, `"amazon"`). Should this be:
```
benbrandt: Definitely my main open question as well.

One option would be that the key in the providers is an arbitrary string and it has a subfield of which type of provider it is compatible with? Potentially more information could then be provided on additional requirements?

If we go this route though, it feels less like a capability and more like something that comes back on the top-level intiailize response.

But this would allow for agents that have potentially multiple openai compatible providers from an api layer, but maybe not all should use the same endpoint (i.e. they support multiple providers, several of which are openai compatible, and you might want to be able to configure all of them)

This opens a lot of UX questions in terms of how these are chosen by the user... but might make things more flexible

docs/rfds/custom-llm-endpoint.mdx
```diff
@@ -0,0 +1,292 @@
...
+
+When a custom endpoint is provided, it may only support a subset of models. For example, a self-hosted vLLM server might only have `llama-3-70b` available, while the agent normally advertises `claude-3-opus`, `gpt-4`, etc.
+
+## Frequently asked questions
+
+> What questions have arisen over the course of authoring this document?
+
+### Why not pass endpoints in the `initialize` request?
+
+Passing endpoints directly in `initialize` would require the client to have already resolved credentials and configured endpoints before knowing whether the agent supports this feature. In practice, the client needs to inspect the agent's capabilities first to decide its authorization strategy — for example, whether to route through a corporate proxy or use direct credentials. A dedicated method after initialization solves this chicken-and-egg problem and keeps capability negotiation separate from endpoint configuration.
```
benbrandt: Yeah overall I like this approach a lot

benbrandt (commented):
Overall I really like this! Some questions but no major objections

docs/rfds/custom-llm-endpoint.mdx
anna239: yes, that was the idea — some agents use different providers for different subagents

docs/rfds/custom-llm-endpoint.mdx
xtmq: Hello Ben, thank you for the blazingly fast review! 

> Is the goal here to have "api compatibility"?
> i.e. I see anthropic and therefore know that I can provide any anthropic compatible endpoint

Yes

> this could become a huge list, and I wonder if this should be driven more by endpoint compatibility?

Yes! this is the most tricky part of this RFD for me also... We can also use `string` type here instead of `enum`, and provide common providers in a documentation and comments instead. But it is important to have an agreement between an agent and a client how do they call "anthropic" provider. Otherwise the whole idea of standartization will be broken.

docs/rfds/custom-llm-endpoint.mdx
xtmq: Yes, I will do!

docs/rfds/custom-llm-endpoint.mdx
xtmq: Yes

docs/rfds/custom-llm-endpoint.mdx
xtmq: Yes, because a client may or **may not** provide a custom endpoint. This is optional and up to the client. So the agent should not relay on that information.

docs/rfds/custom-llm-endpoint.mdx
benbrandt: RIght, I guess I was still thinking we have an enum that is a sub property of the nested object...

I am stil not sure if this is a good idea to be honest... this is a tricky one. but I think this does need to be more than some generic input, we need a semantically meaningful definition of a provider in the protocol for sure

benbrandt: I guess there is also a different need of :
- This is an openai compatible provider and
- This is a provider where I expect specific models to be available, but behind a proxy or some other company specific need

xtmq: The second case can be achived by passing custom metadata (model list?) with the provider name. We can even make a protocol extension there to define API when we will understand the needs.

xtmq: For me, the main question right now is how to define provider types so that both the client and the server can be sure they’re referring to the same thing, while still allowing this list of providers to be easily extended in the future without breaking compatibility in both agents and clients.

docs/rfds/custom-llm-endpoint.mdx
xtmq: That's a great idea. The only downside I see here is the potential for overengineering. We still don't really know how agents and clients will want to use different providers for the same protocol, or whether they’ll want to at all.

Maybe it’s worth considering an approach where the setLlmEndpoints request could accept multiple URLs with different conditions instead. 

But I would vote for simplicity right now.

cdxiaodong:
nice sir！！！

IceyLiu:
good PR, we do need it

docs/rfds/custom-llm-endpoint.mdx
xtmq: I have updated the PR to address provider issues! I renamed LlmProvider to LlmProtocol and made it `string`

docs/rfds/custom-llm-endpoint.mdx
xtmq: I suggested a possible solution

docs/rfds/custom-llm-endpoint.mdx
```diff
@@ -0,0 +1,327 @@
...
+
+The protocol wire format uses plain strings to keep the specification stable and avoid blocking adoption of new LLM APIs. A fixed enum would require a spec update every time a new protocol emerges.
+
+Instead, well-known protocol identifiers are provided by SDKs as convenience constants:
+
+**TypeScript** — an open string type with predefined values:
+```typescript
+type LlmProtocol = string & {};
+
+const LlmProtocols = {
```
xtmq: @benbrandt is it possible to provide such well-known constants in SDK?

leovigna:
I like the concept of client-side LLM providers. 
Regarding config isn't session-config with a "provider" category sufficient for this though? https://agentclientprotocol.com/protocol/session-config-options#session-config-options
I suppose the main limitation is that the spec for config only defines `string` values at the moment and doesn't have the option to add multiple values

xtmq:
> I like the concept of client-side LLM providers. Regarding config isn't session-config with a "provider" category sufficient for this though? https://agentclientprotocol.com/protocol/session-config-options#session-config-options I suppose the main limitation is that the spec for config only defines `string` values at the moment and doesn't have the option to add multiple values

I replied in the FAQ section

docs/rfds/custom-llm-endpoint.mdx
```diff
@@ -0,0 +1,386 @@
...
+```
+
+**providers/set Request:**
+
+```json
+{
+  "jsonrpc": "2.0",
+  "id": 2,
+  "method": "providers/set",
+  "params": {
```
anna239, line 273: I think having it like this 
```
"params": {
    "id": "main",
   "current": {
 "apiType": "anthropic",
    "baseUrl": "https://llm-gateway.corp.example.com/anthropic/v1",
    "headers": {
      "X-Request-Source": "my-ide"
    }
}
  }
```

would be better because we'll be able to use the same objects for parsing providers/list and setting a new value. In this case we can also get rid of `disable` method, as you can just invoke 
```
"params": {
    "id": "main",
   "current": null
  }
```

xtmq: Yes, you are right! But this API urges agents to expose their headers (with possible tokens) to a client.

anna239: Why is that? I mean you can still skip headers when you return list of the providers

benbrandt (approved):
I like the current state a lot! 
Small question:

What do you think of `providers/unset` instead of disable?

Just so it matches providers/set and seems to match the semantic of setting to null?

Just an idea

Then there is the possibility of if it is required that it would "unset" back to the default value?

I guess I just wonder if there is a helpful distinction of knowing if this provider has been "edited" or not and you might want to go back to defaults

benbrandt:
Let me clarify, I guess what i am asking is new: It seems we need a way to 

set - override settings
unset/disable - turn off a provider
reset - go back to "defaults" for a given one

Does that make sense to you?

xtmq:
Yes! it make sense. But having 3 similar commands: set/unset/reset look a bit confusing from the first glance... Maybe it is fine..

xtmq:
@benbrandt I want to preserve "disable":

`disable` explicitly describes runtime behavior: the provider must not be used for LLM calls and is represented in `providers/list` as `current: null`.

`unset` is ambiguous in this API because it can mean either "disable this provider" or "remove client override and return to default".

About the "reset" - nice thing to add in the future. Right now we specify these providers per agent process and do not store them in the file system. Looks like it is not urgent. Let's keep the idea for the future.

benbrandt:
Fair. Yeah I guess reset would be nice. Since I htink some agents might store this config. but fine for now

lifeizhou-ap:
Hi @xtmq, @anna239, and @benbrandt, 

I have a question about how `ProviderInfo` is intended to be used.

Currently goose exposes pre-session provider discovery and configuration through goose-specific ACP methods. We would like to standardize the common parts in ACP so other agents and clients can support the same flows without goose-specific APIs.

Goose Desktop currently has two distinct provider flows:

- For an active session, it already uses `session/set_config_option` to switch the provider and model.
- Without a session, goose discovers provider integrations such as Anthropic, OpenAI, or Ollama, and custom providers shows their setup requirements, configures or authenticates them, and saves a default.

We also have an agent/persona builder use case. Here, “agent/persona” means a reusable, client-managed persona configuration, not the ACP agent process or an ACP session.

When creating one, the client needs to:
- Discover the available provider integrations.
- Let the user choose one.
- Store that provider selection with the persona.

This selection is part of the persona’s saved definition, not the state of an active ACP session.

### Question
The RFD uses both `"main"` and `"openai"` as provider IDs, so we are unsure what `ProviderInfo` is intended to represent:

1. An agent-defined routing target such as `"main"`, whose endpoint the client can override.
2. A provider integration available to the agent—such as Anthropic, OpenAI, Ollama, or a user-defined custom provider—that clients can display, configure, and reference from an agent/persona.

Can `ProviderInfo` represent either of these, or is the current RFD limited to routing targets?

If it is routing-only, would pre-session provider discovery and setup be an extension of this RFD, or should it be a separate proposal?

PR #662, opened 2026-03-06
docs(rfd): Symmetric filesystem capabilities

codefromthecrypt:
Proposes that `writeTextFile: true` implies `readTextFile: true` in client capabilities. Across 23 open-source clients, none send an asymmetric combination. The unused combinations create unnecessary complexity for agent developers and a buffer consistency question that has no practical use case.

See the [goose discussion](https://github.com/block/goose/pull/7668#discussion_r2889311685) for context on the buffer consistency problem.

### DIFF
--- a/docs/rfds/symmetric-fs-capabilities.mdx
+++ b/docs/rfds/symmetric-fs-capabilities.mdx
@@ -1,0 +1,113 @@
+---
+title: "Symmetric filesystem capabilities"
+---
+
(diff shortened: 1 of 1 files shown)


benbrandt:
Given the decision to remove these capabilities from the client in v2, going to close: https://github.com/agentclientprotocol/agent-client-protocol/pull/1346

PR #712, opened 2026-03-10
Add Mobvibe to mobile clients

Eric-Song-Nop:
## Summary
- add Mobvibe to the ACP clients page under `Mobile clients`
- link both the product site and GitHub repository
- list current platform support as Android and Web only

## Mobvibe
Mobvibe is a mobile-first client for interacting with coding agents remotely, and this adds it to the public ACP client directory alongside similar mobile clients such as Happy.


ignatov:
It was merged today, please see the client page

PR #718, opened 2026-03-10
Allow non-UUIDs as message ids

SteffenDE:
PR #244 introduced the `messageId` field and required both clients and agents to use the UUID format for IDs. It also states that the ID should be treated as an opaque value.

I propose to drop the requirement on UUIDs and instead focus on IDs being unique with the exact format not being imposed by ACP.

For example, OpenCode uses a different message ID format internally and I'm not sure if it would be worth the effort to add logic for using UUIDs just for ACP.

### DIFF
--- a/docs/rfds/message-id.mdx
+++ b/docs/rfds/message-id.mdx
@@ -147,7 +147,7 @@ The `messageId` field would be:
 
 - **Optional** on `agent_message_chunk`, `user_message_chunk`, `agent_thought_chunk` updates and `session/prompt` requests (as `messageId`)
 - **Optional** in `session/prompt` responses (as `userMessageId`)
-- **UUID format** - Both clients and agents MUST use UUID format to ensure collision avoidance
+- **Uniqueness** - Both clients and agents must ensure that generated IDs are unique per session, for example by using UUIDv4/v7
 - **Unique per message** within a session
 - **Stable across chunks** - all chunks belonging to the same message share the same `messageId`
 - **Opaque** - Implementations treat it as an identifier without parsing its structure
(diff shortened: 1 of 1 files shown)


benbrandt:
I kind of had this concern before. But now I also have a concern: the RFD as it stands allows for BOTH sides to generate ids.

If it was only one side, I don't think we'd need to standardize. This was also my argument for potentially starting with just agent-generated ids.

I wonder if the agent should perhaps emit a notification for a user message once it receives it to work around this issue?

SteffenDE:
That would work for me.

Another problem I encountered when looking into message ids for Claude is that the Claude Agent SDK does not emit the ID for the agent message until it is fully generated. So if we'd want to support partial message streaming and message IDs in claude-agent-acp, we'd need to have a way for ACP to generate a temporary message ID for agent chunks and then send an update with the "final" ID later. (Assuming we cannot get Anthropic to change the SDK to provide stable IDs for stream events. Probably very unlikely)

benbrandt:
Yeah ok this was kind of my concern from the beginning... message ids are very useful. Probably any agent who natively supports ACP will be able to do this, but Claude Code is always the most popular but we're limited because it is an adapter...

SteffenDE:
fwiw we don't even need a notification. The RFD already says that the agent must provide the `userMessageId` on the response when it generated the ID, so we could just not include user provided IDs for now.

yordis (commented):
I support this change, enforcing a UUID format other than `string` is already limiting the system design without knowing the situation.

I would change the language to `SHOULD`

thezzisu:
Personally, I think it may be better to be "ID **MUST** be treated as an opaque value" and do not require a format, which allows both agents and clients using their own format for message ID, with only one restriction: the ID must be unique.

A common use case: Clients may batching multiple messages (user message, system notifications) into one prompt turn (since ACP prompt turn can only send one message). How to distinguish these messages can be achieved using a special "uuid1:range1;uuid2:range2" message ID which would enable clients to identify & unpack these messages.

btw the zulipchat seems to not having a topic for message ID under #rfds

benbrandt:
Given the issues here, and the move for completely agent-owned message ids in v2, I simplified the RFD and removed the UUID restriction: https://github.com/agentclientprotocol/agent-client-protocol/pull/1344

Going to close this one

PR #721, opened 2026-03-10
Add RFD: Streamable HTTP & WebSocket Transport

alexhancock:
Adding an RFD for a streamable HTTP and websocket transport

### DIFF
--- a/docs/rfds/streamable-http-websocket-transport.mdx
+++ b/docs/rfds/streamable-http-websocket-transport.mdx
@@ -36,9 +36,9 @@ ACP only has stdio (inherited from MCP). There is no standard remote transport, 
 
 Follows the MCP 2025-11-25 Streamable HTTP spec with these adaptations:
 
-- Connection header: `Acp-Connection-Id` (replaces `Mcp-Session-Id`)
+- Connection header: `Acp-Connection-Id` (similar to `Mcp-Session-Id`)
 - Session header: `Acp-Session-Id` (new, no MCP equivalent)
-- Protocol version header: `Acp-Protocol-Version` (not `MCP-Protocol-Version`)
+- Protocol version header: `Acp-Protocol-Version`
 - Endpoint path: conventionally `/acp`
 
 ### 2. Separates connection identity from session identity with two headers
@@ -71,7 +71,7 @@ Clients MUST accept, store, and return cookies set by the server on all HTTP-bas
 
 ### 6. Preserves the full ACP lifecycle
 
-The `initialize` → `initialized` → messages → close lifecycle is identical regardless of transport. The `Acp-Connection-Id` header binds requests to the initialized connection and its negotiated capabilities. The `Acp-Session-Id` header (introduced after `session/new`) identifies the active ACP session. Both are transport-agnostic concepts.
+The `initialize` → `initialized` → messages → close lifecycle is identical regardless of transport. The `Acp-Connection-Id` header binds requests to the initialized connection and its negotiated capabilities. The `Acp-Session-Id` header (introduced after `session/new`) identifies an active ACP session.
 
 ## Shiny future
 
@@ -128,7 +128,7 @@ Client                              Server
   │─── POST /acp ─────────────────────>│  { method: "initialize", id: 1 }
   │    Accept: application/json,       │  (no Acp-Connection-Id header)
   │            text/event-stream       │
-  │              ┌─────────────────────│  Server creates connection, opens SSE stream
+  │              ┌─────────────────────│  Server creates connection
   │              │ (SSE stream open)   │
   │<─────────────│─ SSE event ─────────│  { id: 1, result: { capabilities } }
   │              │                     │  Response includes Acp-Connection-Id header
@@ -377,3 +377,4 @@ Clients SHOULD include it on all requests after initialization. Not yet implemen
 
 - **2025-03-10**: Initial draft based on the RFC template and goose reference implementation.
 - **2026-04-01**: Introduced a two-header identity model: `Acp-Connection-Id` (returned at `initialize`, binds to the connection) and `Acp-Session-Id` (returned at `session/new`, scopes to a session). This addresses feedback that the original single `Acp-Session-Id` conflated transport binding with ACP session identity, and enables session-scoped GET listener streams for targeted server-to-client event delivery. Removed connection-scoped GET streams — all GET SSE listeners now require both `Acp-Connection-Id` and `Acp-Session-Id`.
+- **2026-04-15**: Minor edits
--- a/docs/rfds/streamable-http-websocket-transport.mdx
+++ b/docs/rfds/streamable-http-websocket-transport.mdx
@@ -32,36 +32,46 @@ ACP only has stdio (inherited from MCP). There is no standard remote transport, 
 
 > What are you proposing to improve the situation?
 
-### 1. Adopts MCP Streamable HTTP semantics with ACP-specific headers
+### 1. Mostly adopts MCP Streamable HTTP semantics with ACP-specific headers
 
 Follows the MCP 2025-11-25 Streamable HTTP spec with these adaptations:
 
-- Session header: `Acp-Session-Id` (not `MCP-Session-Id`)
+- Connection header: `Acp-Connection-Id` (replaces `Mcp-Session-Id`)
+- Session header: `Acp-Session-Id` (new, no MCP equivalent)
 - Protocol version header: `Acp-Protocol-Version` (not `MCP-Protocol-Version`)
 - Endpoint path: conventionally `/acp`
 
-### 2. Adds WebSocket as a first-class upgrade on the same endpoint
+### 2. Separates connection identity from session identity with two headers
 
-A GET with `Upgrade: websocket` upgrades to a persistent bidirectional channel — same endpoint, same session model.
+ACP introduces two HTTP headers that together identify the full request context:
 
-This is important for ACP, as its more bidirectional in its nature as a protocol
+- **`Acp-Connection-Id`** (HTTP header) — A transport-level identifier returned by the server in the `initialize` response. It binds all subsequent HTTP requests to the initialized connection and its negotiated capabilities. This is analogous to `Mcp-Session-Id` in MCP Streamable HTTP. Required on all requests after `initialize`.
+- **`Acp-Session-Id`** (HTTP header) — A session-level identifier returned by the server in the `session/new` response (both in the `Acp-Session-Id` response header and the JSON-RPC result body). It identifies a specific conversation. Clients MUST include it on all subsequent requests that operate within a session (e.g., `session/prompt`, `session/cancel`). The same value appears in JSON-RPC params as `sessionId` for methods that require it.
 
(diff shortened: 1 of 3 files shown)


alexhancock:
cc @jh-block @benbrandt

alpgul:
A [project](https://github.com/marimo-team/stdio-to-ws) has already been developed for this feature, and I believe it would be better to keep it as a separate project to avoid increasing code maintenance overhead.

docs/rfds/streamable-http-websocket-transport.mdx
```diff
@@ -0,0 +1,317 @@
...
+  │                                   │
+  │  ═══ Session Initialization ═══   │
+  │                                   │
+  │─── POST /acp ─────────────────────>│  { method: "initialize", id: 1 }
+  │    Accept: application/json,       │  (no Acp-Session-Id header)
+  │            text/event-stream       │
+  │              ┌─────────────────────│  Server creates session, opens SSE stream
+  │              │ (SSE stream open)   │
+  │<─────────────│─ SSE event ─────────│  { id: 1, result: { capabilities } }
+  │              │                     │  Response includes Acp-Session-Id header
```
agentcooper: From what I understand `Acp-Session-Id` here is a transport specific ID. It is a confusing name because ACP session ID is expected to be created later on `session/new` (https://agentclientprotocol.com/protocol/session-setup#session-id).

alexhancock: This is a good point, thanks for catching.

Do you think it's sane to drop the transport specific id and have no `Acp-Session-Id` included in requests unless a `session/new` has been performed? Also interested in alternative resolutions.

mikhailshilkov: > unless a session/new has been performed

`session/prompt` has a `session_id` parameter in its request, which seems to be a separate session ID (actual session ID, not tranport ID). It has to be different from `Acp-Session-Id` because the same `Acp-Session-Id` can call `session/new` multiple times, if I understand it correctly. I'm not sure how exactly `Acp-Session-Id` is supposed to be used in the proposal.

anna239: > Do you think it's sane to drop the transport specific id and have no Acp-Session-Id included in requests

This won't work, as capabilities defined during the initialize phase, so when the client sends their session/new request, the server should know which this Session id. But I agree that the name is confusing here.

anna239:
@alexhancock do I understand it correctly that you suggest opening multiple SSE (first on `initialize` request, then on `session/new` and then on `session/prompt`)? When are those streams closed and how does the agent `decide` which stream does some update belong to?

alexhancock:
> @alexhancock do I understand it correctly that you suggest opening multiple SSE (first on `initialize` request, then on `session/new` and then on `session/prompt`)?

Intent is to follow the request patterns of MCP's streamable http transport https://modelcontextprotocol.io/specification/2025-03-26/basic/transports#sending-messages-to-the-server. From that spec:

"Every JSON-RPC message sent from the client MUST be a new HTTP POST request to the MCP endpoint"

> When are those streams closed

"After all JSON-RPC responses have been sent, the server SHOULD close the SSE stream."

> how does the agent decide which stream does some update belong to?

I don't have a perfect answer to this off the top of my head, but it likely relates to the discussion [here](https://github.com/agentclientprotocol/agent-client-protocol/pull/721#discussion_r2993277282) about session ids. 

Should there perhaps be a session id which only corresponds to the concept of ACP Session ID (nothing transport level) and then a request id to correspond to one client request within a session?

Open to ideas!

anna239:
@alexhancock 
> "After all JSON-RPC responses have been sent, the server SHOULD close the SSE stream."

does it mean, that f.i. the first SSE stream that was opened for `initialize` request will be closed once `initialize` response is streamed? I think this won't work as the agent can send some data to the client after `initialize` response but before `session/new` is called

alexhancock:
> @alexhancock
> 
> > "After all JSON-RPC responses have been sent, the server SHOULD close the SSE stream."
> 
> does it mean, that f.i. the first SSE stream that was opened for `initialize` request will be closed once `initialize` response is streamed? I think this won't work as the agent can send some data to the client after `initialize` response but before `session/new` is called

The way I had been thinking about it, each SSE stream opened by a POST request closes after its response is sent. But the agent can still reach the client between requests using a dedicated GET listener stream, similar to this part of MCP's specification of its streamable HTTP transport (https://modelcontextprotocol.io/specification/2025-03-26/basic/transports#listening-for-messages-from-the-server)

> The client MAY issue an HTTP GET to the MCP endpoint. This can be used to open an SSE stream, allowing the server to communicate to the client, without the client first sending data via HTTP POST.

For the specific case you mentioned, if the agent needs to send data after `initialize` but before `session/new`, the client could receive this if it has the GET-initiated listener stream open. The server would send messages on that stream, not on the POST-initiated stream from the `initialize` request.

**However**, I'm wondering if we should consider a simpler alternative that's more aligned with idiomatic SSE usage:

**Option 1 (current proposal)**: Multiple streams as described above and in the current draft of the RFD
- Request-scoped POST streams that close after response
- Optional GET listener stream for out-of-band messages
- Follows MCP's pattern closely

**Option 2 (alternative)**: Single long-lived stream
- Client opens one GET stream at session start, keeps it open
- ALL server→client messages flow on this stream
- POSTs return `202 Accepted` and close their response instead of opening streams
- Simpler mental model, more idiomatic SSE

The second option would be a deviation from MCP's spec, but might be better suited to ACP's more bidirectional nature and the high volume of notifications from long-running prompts. It would also naturally solve the concern you raised about messages between requests. This option would likely need to rely more heavily on cookie-based session affinity.

I'm curious what you and others think - does the multi-stream approach make sense for ACP, or would a single long-lived stream be simpler and more practical? Open to ideas and feedback.

Also agree we need to clean up the semantics of `Acp-Session-Id` generally in this proposal and get to clarity there. Also open to ideas/approaches to that.

alexhancock:
@agentcooper @mikhailshilkov @anna239 

I have pushed a new revision of the doc to address feedback.

1. It now proposes two concepts, an `Acp-Connection-Id` header to model a client/server connection that has been initialized with capabilities per [this discussion](https://github.com/agentclientprotocol/agent-client-protocol/pull/721#discussion_r2983110892).
2. It clarifies how streams will work (both per `POST` and single session specific long-running `GET`) per [this discussion](https://github.com/agentclientprotocol/agent-client-protocol/pull/721#issuecomment-4140407708).

I would really like opinions/feedback on these resolutions to the original issues from the first draft, and any alternative ideas if you see opportunities to improve the proposal. Thank you!

cc @jh-block @benbrandt @DOsinga

anna239:
@alexhancock thank you, current proposal looks good! We have some protocol changes we're planning for some time already, they're related to how ACP works with prompt request. We'd like to review those and see if it makes sense to deliver them before HTTP implementation, I'm sorry for the delay, we need several more days for it

qas:
Hey @alexhancock, nice RFD — this is the piece I've been waiting for.

I'm building [ageneral.ai](https://ageneral.ai) — cloud-hosted agents on remote VMs, IDE clients connecting through a reverse proxy. We're building an ACP-to-remote-agent bridge (stdio on the IDE side, WebSocket through a control plane proxy to the agent) and this RFD directly affects our design, so wanted to share some thoughts.

**On Option 1 vs Option 2:** +1 for Option 2 (single long-lived stream). Our existing agent-to-gateway connections use a single WebSocket for the entire session — all notifications, tool calls, streaming chunks on one connection. The "which stream does this belong to" problem @anna239 raised just doesn't exist with a single persistent connection. ACP sessions are long-lived and notification-heavy, so this fits much better than managing multiple POST+SSE streams.

**Happy to see WebSocket as a required client capability.** That's the right call. For cloud-hosted agents behind a reverse proxy, WebSocket is the natural transport — single connection, full-duplex, no stream routing or sticky session complexity. Proxies/LBs (Caddy, nginx, Envoy) all handle WS upgrades natively. The Streamable HTTP fallback for serverless makes sense too.

**One gap: the reverse proxy / gateway pattern.** In a cloud setup, clients don't connect directly to the agent — they hit an API gateway that authenticates the user and proxies to the right host. A few things worth considering:

- `Acp-Connection-Id` needs to survive through proxies. With WebSocket this is free (one connection = one identity). With Streamable HTTP the proxy has to correlate multiple POSTs to the same logical connection, which is harder — the cookie requirement helps but might not be enough if the proxy is a different origin than the agent.
- Transport-level auth (JWT in a header or query param on the upgrade request) often happens *before* ACP `authenticate`. Cloud deployments typically verify the user at the transport layer before any JSON-RPC flows. Might be worth a note in the spec that transport-level auth is separate from and may precede ACP-level auth.
- Split transports are real — client↔proxy can be WebSocket while proxy↔agent is stdio. The proxy-chains RFD touches on this but it'd be good for the transport RFD to explicitly acknowledge that transports can be layered (the proxy terminates one transport and speaks another to the agent).

**Connection ID vs Session ID split is great.** The two-header model makes the lifecycle clear — connection is ephemeral (scoped to one `initialize`), session is durable (survives reconnects via `session/load`). This is exactly what we need for the "disconnect, reconnect on a new connection, load the same session" flow.

Happy to go deeper on any of this.

benbrandt (approved):
I think this looks great for Draft stage thanks so much! 
Let's start trying this out and see what we need to change once the rubber meets the road 😄

johnhenry:
Late comment on a merged RFD, but wanted to leave a concrete data point here since this doc's language ("SSE `Last-Event-ID`-style resumption lets a reconnecting client replay messages missed while disconnected") is exactly the gap I've hit building a client on top of `@agentclientprotocol/sdk`.

I maintain [acp-query](https://github.com/johnhenry/acp-query) (`@johnhenry/acpq`), a reactive session/turn store for ACP. Its honest current position on reconnection, from its own docs:

> ACP has no replay. Session streams cannot be resumed with history, and there is no full read to reconcile against — `session/update` deltas are meaningful only in arrival order and the agent never re-sends them. A reconnect therefore leaves session state gappy [...] recovery is a fresh session or a re-prompt.

`session/resume` + `replayFrom` (v2 draft) is real progress here, but as currently defined `ReplayFrom` only has one concrete variant — `{"type": "start"}` (replay everything) — with the rest of the union reserved as an open extension point for "custom or future replay cursor" types. Meanwhile v2 also just made `messageId` **required** on `agent_message`/`user_message`/`agent_message_chunk`/`agent_thought_chunk` (per the messageId RFD). That seems like exactly the missing piece: a second `ReplayFrom` variant keyed on `messageId` — `{"type": "messageId", "messageId": "..."}` or similar — would give clients a concrete, cheap way to say "replay from here" instead of "replay everything or nothing," using an identifier the protocol already guarantees exists on every message going forward.

Not asking for a new wire concept, just for the extension point `ReplayFrom` already reserves to get filled in with the piece `messageId` now makes possible. Cross-referencing https://github.com/orgs/agentclientprotocol/discussions/396, which asked almost this exact question with no resolution on record — and I'll also open a Protocol Suggestions discussion so this has a live thread now that this PR is merged.

PR #723, opened 2026-03-11
Fix remnant stop

SteffenDE:
Relates to #701.


benbrandt:
Whoops just fixed in #724

PR #754, opened 2026-03-15
feat(unstable): initial unstable nes implementation

anna239:
Initial implementation for NES support

### DIFF
--- a/src/nes.rs
+++ b/src/nes.rs
@@ -646,10 +646,10 @@ pub struct ClientNesCapabilities {
     pub jump: Option<NesJumpCapabilities>,
     /// Whether the client supports the `rename` suggestion kind.
     #[serde(skip_serializing_if = "Option::is_none")]
-    pub rename: Option<NesRenameActionCapabilities>,
+    pub rename: Option<NesRenameCapabilities>,
     /// Whether the client supports the `searchAndReplace` suggestion kind.
     #[serde(skip_serializing_if = "Option::is_none")]
-    pub search_and_replace: Option<NesSearchAndReplaceActionCapabilities>,
+    pub search_and_replace: Option<NesSearchAndReplaceCapabilities>,
     /// The _meta property is reserved by ACP to allow clients and agents to attach additional
     /// metadata to their interactions. Implementations MUST NOT make assumptions about values at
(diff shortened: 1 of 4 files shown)


src/nes.rs
```diff
@@ -0,0 +1,2154 @@
...
+#[serde(rename_all = "camelCase")]
+#[non_exhaustive]
+pub struct NesActionSuggestion {
+    /// Unique identifier for accept/reject tracking.
+    pub id: String,
+    /// The IDE action to perform (must match a client-advertised action).
+    pub action_id: String,
+    /// Action parameters matching the schema from the client's capability.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub arguments: Option<serde_json::Value>,
```
benbrandt: are the arguments really an arbitrary value? How does the caller know what to put in there?

src/nes.rs
```diff
@@ -0,0 +1,2154 @@
...
+}
+
+/// Capabilities for `document/didChange` events.
+#[derive(Default, Debug, Clone, Serialize, Deserialize, JsonSchema, PartialEq, Eq)]
+#[serde(rename_all = "camelCase")]
+#[non_exhaustive]
+pub struct NesDocumentDidChangeCapabilities {
+    /// The sync kind the agent wants: `"full"` or `"incremental"`.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub sync_kind: Option<TextDocumentSyncKind>,
```
benbrandt: I think this sync kind type either needs a default value or we need to require it?

src/nes.rs
```diff
@@ -0,0 +1,2154 @@
...
+#[non_exhaustive]
+pub struct NesUserAction {
+    /// The kind of action (e.g., "insertChar", "cursorMovement").
+    pub action: String,
+    /// The URI of the file where the action occurred.
+    pub uri: String,
+    /// The line number where the action occurred.
+    pub line: u32,
+    /// The character offset where the action occurred.
+    pub offset: u32,
```
benbrandt: Should this be a position?

benbrandt:
Do we want a "nes session close" method of some kind?

PR #792, opened 2026-03-19
feat(unstable): elicitation for session, tool call, and requests

anna239:
… session scope

### DIFF
--- a/src/elicitation.rs
+++ b/src/elicitation.rs
@@ -1057,7 +1057,7 @@ impl CreateElicitationResponse {
 /// The user's action in response to an elicitation.
 #[derive(Debug, Clone, Serialize, Deserialize, JsonSchema, PartialEq)]
 #[serde(tag = "action", rename_all = "snake_case")]
-#[schemars(inline, extend("discriminator" = {"propertyName": "action"}))]
+#[schemars(extend("discriminator" = {"propertyName": "action"}))]
 #[non_exhaustive]
 pub enum ElicitationAction {
     /// The user accepted and provided content.
@@ -1074,7 +1074,6 @@ pub enum ElicitationAction {
 ///
 /// The user accepted the elicitation and provided content.
 #[derive(Debug, Clone, Serialize, Deserialize, JsonSchema, PartialEq)]
-#[schemars(inline)]
 #[serde(rename_all = "camelCase")]
 #[non_exhaustive]
 pub struct ElicitationAcceptAction {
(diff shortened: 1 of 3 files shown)


yordis:
@anna239 could you share a bit the behind the scenes conversations to understand the decision making around it? 
I can not tell what led to the changes, to be able to understand what we missed and what we need to adjust in our end as well

src/elicitation.rs
```diff
@@ -870,13 +870,18 @@ impl ElicitationUrlCapabilities {
...
 #[non_exhaustive]
 pub struct ElicitationRequest {
-    /// The session ID for this request.
-    pub session_id: SessionId,
+    /// Optional session ID if this elicitation is tied to a specific session.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub session_id: Option<SessionId>,
+    /// Optional request ID if this elicitation is tied to a specific request.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub request_id: Option<RequestId>,
```
benbrandt: Do we want to require one or the other? I can help turn this into an enum to express that you have to have one (and not non or both?)

benbrandt:
@yordis it's clear that there are use-cases for other methods that this can be helpful. The idea is to have it either for a session or some other request, and then the client would need to have some way to express general elicitations, for the auth/configuration phases to be concrete

yordis:
> for the auth/configuration phases to be concrete

Related to https://github.com/agentclientprotocol/agent-client-protocol/pull/376#discussion_r2784416804 

@benbrandt that was a precise question in the RFD, and the idea was that such workflow didn't belong to the Elicitation per se, which it is fine, I do not have strong opinions, but overlapping concepts may arise. 

Documenting the decision making would be amazing, so we all have a trail history of the "why" behind it. Sometimes isn't obvious what the motivation is, so it helps to share the intent.

benbrandt:
Sure thing. before we merge this we would need to update the RFD which is where we'd capture that

src/elicitation.rs
anna239: I'm not sure, I think there are valid use cases for `none` option — like you can have a request that is not tied to a particular session or to the json rpc request, the agent can just ask you to provide some additional settings f.i. before any session is started. 
Providing both might be ambiguous (in case you can tide requestId to a session already). 
But maybe I'm overthinking and we should go with either-or

anna239:
@benbrandt what do you think about ability to provide a related tool call id as it's done for request_permissions? I was thinking about combining this with MCP elicitations — what if an agent gets elicitation request while processing doing an MCP tool call — this way it can redirect the request to the user and also provide a tool call id where it happened. wdyt? 
But this can also be done in an update

benbrandt:
Ok so we'd have three variants?
- session elicitation
- tool call elicitation (session_id + tool_call_id)
- global elicitation (request_id)

Seems fine

anna239:
@benbrandt updated code and rfd with `toolCallId`

PR #808, opened 2026-03-21
docs(rfd): client-owned fs/apply_patch

leovigna:
## Summary

This adds an initial RFD draft for a client-owned `fs/apply_patch` capability.

The draft proposes:
- a new client capability under `clientCapabilities.fs.applyPatch`
- a new client-owned method, `fs/apply_patch`
- applying structured multi-file edits against the client's live editor-backed document state, including unsaved buffers
- treating tool-call `diff` content as presentation/reporting, not as a mutation API
- using a V4A-style `apply_patch` envelope as the initial default format, profiled for ACP semantics such as absolute paths

## Why

ACP already has client-owned file access via `fs/read_text_file` and `fs/write_text_file`, but it does not yet have a client-owned mutation method for structured patch-style edits against current buffer state.

That gap matters for coding agents that want precise multi-file edits without falling back to whole-file overwrite.

## Open questions

The draft calls out a few areas to iterate on next:
- the exact ACP profile of the V4A-style grammar
- how preconditions should work in v1
- whether the `patch` field should stay a string envelope or become a richer object
- whether transactional semantics should be required initially

This is intended as an early draft for discussion rather than a finalized wire shape.

### DIFF
--- a/docs/rfds/client-owned-patch-application.mdx
+++ b/docs/rfds/client-owned-patch-application.mdx
@@ -1,0 +1,273 @@
+---
+title: "Client-owned patch application"
+---
+
+Author(s): [@leovigna](https://github.com/leovigna)
+
+## Elevator pitch
+
+> What are you proposing to change?
+
+Add a new client-owned filesystem capability and method, `fs/apply_patch`, that allows an Agent to ask the Client to apply structured multi-file text edits to the Client's current document state, including unsaved editor buffers.
+
+The key idea is:
+
+- ACP should support patch-style mutation as a **Client capability**, not as a tool.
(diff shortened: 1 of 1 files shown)


andymac4182:
Could this be accomplished by providing the tool using MCP over ACP? https://agentclientprotocol.com/rfds/mcp-over-acp

benbrandt:
Yeah my concern here is that the V4A is pretty tied to openai agents and is something I expect to be iterated on. I'm not sure I would want this at the protocol level...

benbrandt:
GIven we will be removing client FS capabilities in v2, going to close this https://github.com/agentclientprotocol/agent-client-protocol/pull/1346

PR #832, opened 2026-03-24
Anna.zhdan/workspance state

anna239:
---
title: "Client Editor State Methods"
---

- Author(s): [@anna239](https://github.com/anna239)

## Elevator pitch

Add three optional client methods that expose the current editor state to the agent:

1. `workspace/open_documents` for opened files
2. `workspace/recent_documents` for recent files
3. `workspace/active_document` for the currently focused file

These methods are intentionally shaped to match VS Code concepts (`workspace.textDocuments`, `window.activeTextEditor`, and VS Code's internal recently-opened model) while preserving ACP conventions and capability-based opt-in. All three methods live under the `workspace/` prefix for consistency.

## Status quo

Today ACP lets an agent read and write files (`fs/read_text_file`, `fs/write_text_file`) when it already knows a path, but there is no standard way to ask:

1. Which files are currently opened in the editor?
2. Which files were recently opened?
3. Which file currently has focus?

This causes practical issues:

1. Agents must infer context from prompt text or external heuristics.
2. Clients add context unconditionally, which can decrease the agent's performance.

Introduce three optional `client` methods and corresponding capabilities:

1. `workspace/open_documents`
2. `workspace/recent_documents`
3. `workspace/active_document`

Agents must check capability flags before calling these methods.

### Capability advertisement

```json
{
  "clientCapabilities": {
    "workspace": {
      "openDocuments": {},
      "recentDocuments": {},
      "activeDocument": {}
    }
  }
}
```

### URI rules

All returned `uri` values **MUST** use the `file:///` scheme with absolute paths.

### Method: `workspace/open_documents`

Returns currently open text documents that map to absolute file paths.

#### Request

```json
{
  "jsonrpc": "2.0",
  "id": 11,
  "method": "workspace/open_documents",
  "params": {
    "sessionId": "sess_abc123def456"
  }
}
```

#### Response

```json
{
  "jsonrpc": "2.0",
  "id": 11,
  "result": {
    "documents": [
      {
        "uri": "file:///Users/alice/dev/acp/src/client.rs",
        "languageId": "rust"
      },
      {
        "uri": "file:///Users/alice/dev/acp/docs/protocol/file-system.mdx",
        "languageId": "markdown"
      }
    ]
  }
}
```

### Method: `workspace/recent_documents`

Returns recently opened files (absolute paths), optionally bounded by `limit`.

#### Request

```json
{
  "jsonrpc": "2.0",
  "id": 12,
  "method": "workspace/recent_documents",
  "params": {
    "sessionId": "sess_abc123def456",
    "limit": 50
  }
}
```

#### Response

```json
{
  "jsonrpc": "2.0",
  "id": 12,
  "result": {
    "documents": [
      {
        "uri": "file:///Users/alice/dev/acp/src/rpc.rs",
        "languageId": "rust"
      },
      {
        "uri": "file:///Users/alice/dev/acp/docs/protocol/session-setup.mdx",
        "languageId": "markdown"
      }
    ]
  }
}
```

### Method: `workspace/active_document`

Returns the currently focused text editor file, or `null` when no file-backed text editor is focused.

#### Request

```json
{
  "jsonrpc": "2.0",
  "id": 13,
  "method": "workspace/active_document",
  "params": {
    "sessionId": "sess_abc123def456"
  }
}
```

#### Response

```json
{
  "jsonrpc": "2.0",
  "id": 13,
  "result": {
    "document": {
      "uri": "file:///Users/alice/dev/acp/src/client.rs",
      "languageId": "rust"
    }
  }
}
```

## Further development

Once this exists:

1. Agents can implement their own logic for context gathering.
2. `workspace/active_document` can be extended with additional fields such as `visibleRange` and cursor `position` to provide richer editor context.

## Frequently asked questions

1. Why `workspace`?

Was looking for something general enough that would work not only for IDEs.

## Revision history

- 2026-03-17: Initial draft.

### DIFF
--- a/docs/rfds/client-editor-state-methods.mdx
+++ b/docs/rfds/client-editor-state-methods.mdx
@@ -1,0 +1,181 @@
+---
+title: "Client Editor State Methods"
+---
+
+- Author(s): [@anna239](https://github.com/anna239)
+
+## Elevator pitch
+
+Add three optional client methods that expose the current editor state to the agent:
+
+1. `workspace/open_documents` for opened files
+2. `workspace/recent_documents` for recent files
+3. `workspace/active_document` for the currently focused file
+
+These methods are intentionally shaped to match VS Code concepts (`workspace.textDocuments`, `window.activeTextEditor`, and VS Code's internal recently-opened model) while preserving ACP conventions and capability-based opt-in. All three methods live under the `workspace/` prefix for consistency.
+
+## Status quo
+
+Today ACP lets an agent read and write files (`fs/read_text_file`, `fs/write_text_file`) when it already knows a path, but there is no standard way to ask:
+
+1. Which files are currently opened in the editor?
+2. Which files were recently opened?
+3. Which file currently has focus?
+
+This causes practical issues:
+
+1. Agents must infer context from prompt text or external heuristics.
+2. Clients add context unconditionally, which can decrease the agent's performance.
+
+## What we propose to do about it
+
+Introduce three optional `client` methods and corresponding capabilities:
+
+1. `workspace/open_documents`
+2. `workspace/recent_documents`
+3. `workspace/active_document`
+
+Agents must check capability flags before calling these methods.
+
+### Capability advertisement
+
+```json
+{
+  "clientCapabilities": {
+    "workspace": {
+      "openDocuments": {},
+      "recentDocuments": {},
+      "activeDocument": {}
+    }
+  }
+}
+```
+
+### URI rules
+
+All returned `uri` values **MUST** use the `file:///` scheme with absolute paths.
+
+### Method: `workspace/open_documents`
+
+Returns currently open text documents that map to absolute file paths.
+
+#### Request
+
+```json
+{
+  "jsonrpc": "2.0",
+  "id": 11,
+  "method": "workspace/open_documents",
+  "params": {
+    "sessionId": "sess_abc123def456"
+  }
+}
+```
+
+#### Response
+
+```json
+{
+  "jsonrpc": "2.0",
+  "id": 11,
+  "result": {
+    "documents": [
+      {
+        "uri": "file:///Users/alice/dev/acp/src/client.rs",
+        "languageId": "rust"
+      },
+      {
+        "uri": "file:///Users/alice/dev/acp/docs/protocol/file-system.mdx",
+        "languageId": "markdown"
+      }
+    ]
+  }
+}
+```
+
+### Method: `workspace/recent_documents`
+
+Returns recently opened files (absolute paths), optionally bounded by `limit`.
+
+#### Request
+
+```json
+{
+  "jsonrpc": "2.0",
+  "id": 12,
+  "method": "workspace/recent_documents",
+  "params": {
+    "sessionId": "sess_abc123def456",
+    "limit": 50
+  }
+}
+```
+
+#### Response
+
+```json
+{
+  "jsonrpc": "2.0",
+  "id": 12,
+  "result": {
+    "documents": [
+      {
+        "uri": "file:///Users/alice/dev/acp/src/rpc.rs",
+        "languageId": "rust"
+      },
+      {
+        "uri": "file:///Users/alice/dev/acp/docs/protocol/session-setup.mdx",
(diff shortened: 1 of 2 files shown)


kevklam:
I was going to propose the inverse - a way for clients to automatically provide this context/state to the agent in a dedicated channel where the information can be provided by the client on every turn BUT is guaranteed *not* to get snapshotted into turn history.

anna239:
@kevklam I thought about this approach, but I think the agents have very different approaches to context collection, if client has to provide all the data each time, it can end up sending too much data back and forth. So I decided it's better to give an agent a freedom to implement any strategy with client just providing the info on demand

kevklam:
@anna239 - If you're referring to efficiency of minimizing raw bytes transferred between client/agent - I don't think the amount of data sent each time is significant over stdio (or even http). Personally, my bigger concern is managing context size, especially if there is concern about data freshness - for example if an agent needs to call a tool every turn to ensure it has a fresh view of the workspace, then there could be 20 stale copies of workspace information in the chat history?

For reference - I just submitted my own RFD which I think solves the similar idea using 'push' semantics from client side. However my approach allows the client to include whatever context it thinks is relevant, instead of limiting to only a few tightly schematized pieces of metadata. I think this approach allows more flexibility - i.e. different IDEs have different workspace concepts, so VSCode may include some information like
IDE: VSCode
Workspace file: [path]
Workspace roots: [...]
Launch configuration: ....
Opened editor windows: ....
Opened UX panels: ....

XCode might include different information because it has different innate concepts.
IDE: XCode
Projects loaded: .....

https://github.com/agentclientprotocol/agent-client-protocol/pull/874
[Ephemeral IDE context](https://github.com/kevklam/agent-client-protocol/blob/3ccaf5493a7cdc4fe440fa3a6b56a38915808a8b/docs/rfds/ephemeral-ide-context.mdx)

anna239:
@kevklam I believe that it's only agent who "knows" which context is useful. And it's agent's responsibility to congress the context\not store copies of the file in it

kevklam:
@anna239 I'm wondering how you would implement the agent. Do you mean you will have a cheaper LLM call in an orchestrator to choose which context it should include in the main LLM call? My RFD is more targeted to simpler agents that do not have a built-in planning phase.

My concern is mostly that trying to define every IDE context that any client *could* expose would severely limit capabilities. If your intention is to feed this data into an orchestrator, something like the following might be more flexible:
```
{
  "clientCapabilities": {
    "workspace_context_discovery_for_orchestration": {}
  }
}
```
```
Request
{
  "jsonrpc": "2.0",
  "id": 11,
  "method": "workspace/available_context",
  "params": {
    "sessionId": "sess_abc123def456"
  }
}
Response
{
  "jsonrpc": "2.0",
  "id": 11,
  "result": {
    "available_context": [
      {
        "name": "recent_documents",
        "description": "A list of recent documents the user has opened, including path and file size"
      },
      {
        "name": "workspace_tree",
        "description": "The full hierarchical file/folder tree of the workspace"
      },
      {
        "name": "workspace_roots",
        "description": "Just the root folder(s) of the workspace"
      },
      {
        "name": "editor_contents",
        "description": "What's currently visible on screen in the currently active editor"
      }
    ]
  }
}
```
```
Request
{
  "jsonrpc": "2.0",
  "id": 11,
  "method": "workspace/get_context",
  "params": {
    "context": [
       "workspace_tree",
       "editor_contents"
    ]
  }
}
Response
{
  "jsonrpc": "2.0",
  "id": 11,
  "result": {
    "context": [
       { "name": "workspace_tree", "text": "....." },
       { "name": "editor_contents", "text": "....." }
    ]
  }
}
```

docs/rfds/client-editor-state-methods.mdx
```diff
@@ -0,0 +1,182 @@
...
+title: "Client Editor State Methods"
+---
+
+- Author(s): [@anna239](https://github.com/anna239)
+
+## Elevator pitch
+
+Add three optional client methods that expose the current editor state to the agent:
+
+1. `workspace/open_documents` for opened files
```
benbrandt, line 11: @anna239 do you want to revisit this in light of the NES experiments and proposed methods?

PR #838, opened 2026-03-24
feat: additional directories

egor-baranov:
## Summary
- add `additionalDirectories` support to session lifecycle requests, `session/list`, `SessionInfo`, and `SessionCapabilities`
- document multi-root session behavior in the protocol docs and align the session-info update RFD note with immutable workspace roots
- regenerate stable and unstable schemas plus generated schema docs

## Verification
- `npm run generate`
- `npm run check`

### DIFF
--- a/docs/docs.json
+++ b/docs/docs.json
@@ -82,6 +82,9 @@
                 "group": "Draft: In Progress and May Change",
                 "hidden": true,
                 "pages": [
+                  "protocol/draft/session-setup",
+                  "protocol/draft/session-list",
+                  "protocol/draft/file-system",
                   "protocol/draft/cancellation",
                   "protocol/draft/schema"
                 ]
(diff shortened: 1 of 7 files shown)


schema/schema.json
```diff
@@ -1476,6 +1476,13 @@
           "description": "The _meta property is reserved by ACP to allow clients and agents to attach additional\nmetadata to their interactions. Implementations MUST NOT make assumptions about values at\nthese keys.\n\nSee protocol docs: [Extensibility](https://agentclientprotocol.com/protocol/extensibility)",
           "type": ["object", "null"]
         },
+        "additionalDirectories": {
```
anna239: Additional directories are yet unstable, so they should be added to the unstable schema only

egor-baranov:
Re-requesting @anna239 review because of updated nullable schema for additionalDirectories

anna239:
looks good to me

PR #863, opened 2026-03-27
Update auth method terminal

anaslimem:
## Summary

Change `AuthMethodTerminal.env` from `HashMap<String, String>` to `Vec<EnvVariable>` for consistency with `McpServerStdio.env`.

### Code Changes (`src/agent.rs`)

- Changed `AuthMethodTerminal.env` field type from `HashMap<String, String>` to `Vec<EnvVariable>`
- Updated builder method to accept `Vec<EnvVariable>`
- Removed unused `HashMap` import
- Updated test to use new `EnvVariable` struct

### Documentation Updates

- `docs/rfds/auth-methods.mdx`: Updated JSON example and field description
- `docs/protocol/draft/schema.mdx`: Updated `env` type from `object` to `EnvVariable[]`

### Code
```rust
// Before
pub env: HashMap<String, String>

// After  
pub env: Vec<EnvVariable>


benbrandt:
Good point, likely left over from early auth method experiments. 
The problem is we have some large clients in use requiring the current serialization method. We are working on hardening this up for unstable things in the future. But I am not sure this is worth it. We can clean up in the v2 implementation if necessaryu

PR #874, opened 2026-03-29
RFD: Runtime context on session/prompt

kevklam:
## Summary

Adds a draft RFD proposing `runtimeContext` as optional host-provided runtime context on `session/prompt`.

## Proposal

- add `sessionCapabilities.runtimeContext?: {}` as a yes/no capability
- add optional `runtimeContext` to `session/prompt`
- define `runtimeContext` as text-only items shaped like `{ title?: string, text: string }[]`
- make it explicit that `runtimeContext` is used by the agent when constructing the prompt for that turn, but is not part of durable conversation history or session replay

## Motivation

ACP currently has durable session metadata like `cwd` and `additionalDirectories`, and normal prompt content on `session/prompt`, but it does not have a standard way for clients to provide fresh host-side IDE/runtime context for a turn without snapshotting that context into prompt history.

This is especially relevant for IDE-derived context that clients may want to provide freshly on many turns, but that is too large, too expensive, or too low-value to durably record on every turn.

## Design notes

- this proposal standardizes the mechanism and lifecycle semantics, not a fixed schema of IDE facts
- `runtimeContext` is push-based and travels with `session/prompt`
- `runtimeContext` is text-only in this proposal
- clients remain free to decide what runtime context to send for a given turn
- agents should normally incorporate `runtimeContext`, but may trim or omit items if needed for prompt-construction or context-limit reasons

## Non-goals

- replacing `cwd`
- obsoleting `additionalDirectories` as a goal of this RFD
- standardizing a fixed set of IDE fields that agents can take code dependencies on
- recommending fallback behavior when an agent does not support `runtimeContext`

### DIFF
--- a/docs/rfds/ephemeral-ide-context.mdx
+++ b/docs/rfds/ephemeral-ide-context.mdx
@@ -1,0 +1,233 @@
+---
+title: "Ephemeral IDE Context for Sessions"
+---
+
+- Author(s): [kevklam](https://github.com/kevklam)
+
+## Elevator pitch
+
+> What are you proposing to change?
+
+ACP currently gives clients two main ways to provide context to agents during chat-style sessions:
+
+- durable session lifecycle metadata such as `cwd` and `additionalDirectories`; and
+- per-turn prompt content via `session/prompt`, typically snapshotted into conversation history.
+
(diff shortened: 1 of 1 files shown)


benbrandt:
I think without more concrete specification, agents wouldn't be able to do much better than shove it in a user message, since there is no distinction what these items are.

I wonder if #832 or something else along those lines might be better suited for this use case. Closing for now.

PR #883, opened 2026-03-30
fix(rpc): preserve '_' prefix for extension methods and reject empty ext

anaslimem:
## Summary

Fixes extension method handling so custom JSON-RPC methods keep their required `_` prefix across decode/routing/serialization.

## Problem

Extension methods are documented as underscore-prefixed (`_...`), but decode logic stripped `_` and stored the method without it.  
That could cause round-trip drift:

`"_vendor/foo"` -> `"vendor/foo"` (re-emitted)

This breaks extension namespace expectations and risks future method-name collisions.

## Changes

- Preserve extension method names exactly as received on the wire (including `_`) in all decode paths.
- Reject invalid empty extension method name (`"_"`).
- Add regression tests for:
  - request decode prefix preservation (both sides)
  - notification decode prefix preservation (both sides)
  - empty extension method rejection
- Add doc comments clarifying `ExtRequest.method` and `ExtNotification.method` semantics.

### DIFF
--- a/src/ext.rs
+++ b/src/ext.rs
@@ -23,6 +23,9 @@ pub type Meta = serde_json::Map<String, serde_json::Value>;
 #[serde(transparent)]
 #[non_exhaustive]
 pub struct ExtRequest {
+    /// Wire method name for this extension request.
+    ///
+    /// Extension method names must start with `_`, and the name is preserved exactly.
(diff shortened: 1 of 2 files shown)


benbrandt (approved):
thanks!

PR #902, opened 2026-04-02
docs(rfd): Add new plan variants

anna239:


docs/rfds/plan-operations.mdx
```diff
@@ -0,0 +1,157 @@
...
+
+The `id` field on every plan variant identifies a specific plan.
+
+- Each `plan_update` targets the plan with the given `id`
+- Client tracks plans by ID; updates replace the content of that specific plan
+- Different plans may use different types (e.g., one `items` plan and one `markdown` plan)
+
+### Plan Removal
+
+Agent sends a `plan_removed` session update with the plan's `id` to dismiss it. This is a separate session update type from `plan_update`, keeping content updates and lifecycle events distinct.
```
benbrandt, line 61: I wonder if there is a way to `null` out the content like we do in some other areas rather than have another notification type?
Just a thought.

docs/rfds/plan-operations.mdx
```diff
@@ -0,0 +1,157 @@
...
+
+> What are you proposing to improve the situation?
+
+Add a new `SessionUpdate` variant — `plan_update` — that carries a `plan` field. The `plan` field is a tagged union (discriminated by `type`) with the following variants:
+
+| Type       | Description                                           | Specific fields        |
+| ---------- | ----------------------------------------------------- | ---------------------- |
+| `items`    | Structured entries (same semantics as today's `plan`) | `entries: PlanEntry[]` |
+| `file`     | Agent provides a file URI containing the plan         | `uri: string`          |
+| `markdown` | Agent provides raw markdown text                      | `content: string`      |
```
benbrandt, line 33: Can add this later, but is there a need for a label/description for some of these types?

docs/rfds/plan-operations.mdx
```diff
@@ -0,0 +1,157 @@
...
+Item-based (`plan_update`):
+
+```json
+{
+  "sessionUpdate": "plan_update",
+  "plan": {
+    "type": "items",
+    "id": "plan-1",
+    "entries": [
+      { "content": "Step 1", "priority": "high", "status": "pending" }
```
benbrandt, line 93: What is the expectation of the client?
If I get multiple items updates, am I expected to render them all until they are removed?

And I assume the plan is "completed" not by being removed, but by all items being marked as done?

docs/rfds/plan-operations.mdx
anna239: Do you mean like plan.name?

docs/rfds/plan-operations.mdx
anna239: I thought about this option, but then `plan_remove` event would also be typed — meaning that the agent needs to pass the correct entity type (file\markdown\...). So I decided that providing just an id in a separate event is a simpler option. But we can change that

docs/rfds/plan-operations.mdx
anna239: > If I get multiple items updates, am I expected to render them all until they are removed?

Plan update always updates the whole plan, each time you get it you should re-render all the entities. 

> And I assume the plan is "completed" not by being removed, but by all items being marked as done?

yes! Maybe we need something like `plan.status` in general, might be useful for file\markdown plan, but this can be added later

docs/rfds/plan-operations.mdx
anna239: Also not all plan types have content — the one that provides a file name doesn't have content

PR #930, opened 2026-04-08
Add `model_config` rfd

anna239:


benbrandt (approved):
I think this one is easy enough to add as a new category. 
If we want to make this more semantically meaningful we could do some sort of _nested_ config, but it gets tricky fast. 

I think this is easy enough to add as a new category. I do wonder about thought_level though... since it is quite a ubiquitous config, it makes it easier to attach keyboard shortcuts too by having it distinct, but we can also group it. Not feeling strongly about it.

PR #954, opened 2026-04-13
docs(rfd): v2 Prompting

benbrandt:
This is a very early draft with some lingering open questions. But I am
opening now for discussion.

### DIFF
--- a/docs/rfds/v2-prompt.mdx
+++ b/docs/rfds/v2-prompt.mdx
@@ -8,64 +8,51 @@ Author(s): [@benbrandt](https://github.com/benbrandt)
 
 > What are you proposing to change?
 
-For v2 of the protocol wire format, I am proposing that we should move to a fully notification-based format for prompting. In fact, for session updates for both sides in general.
+For v2 of the protocol wire format, I am proposing a change in the lifecycle of the prompt request, allowing for more dynamic session updates from the agent, and unlocking new capabilities in the process.
 
-Once a session is created, both client and agent will be able to update the session at any point in time by sending updates to each other. As I'll go into later, this not only removes some current awkwardness around the prompt request lifecycle, but also provides a more flexible foundation to add features like queued messages and multi-client replay. This can even allow the agent to initiate an interaction in a session rather than requiring it to wait for a user prompt, which is becoming increasingly important for background tasks and agents which may send updates before or after a "turn" is over since its runtime might be different than the main conversation.
+Once a session is created, the agent will be able to send session updates at any point in time, and prompt requests will last until the prompt is accepted, not until the end of the turn. As I'll go into later, this not only removes some current awkwardness around the prompt request lifecycle, but also provides a more flexible foundation to add features like queued messages and multi-client replay. This can even allow the agent to initiate an interaction in a session rather than requiring it to wait for a user prompt, which is becoming increasingly important for background tasks and agents which may send updates before or after a "turn" is over since its runtime might be different than the main conversation.
 
 ## Status quo
 
 > How do things work today and what problems does this cause? Why would we change things?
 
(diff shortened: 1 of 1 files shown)


docs/rfds/v2-prompt.mdx
```diff
@@ -0,0 +1,306 @@
...
+    }
+  }
+}
+```
+
+This would make it a bit more generic, and might be a good alternative.
+
+### What alternative approaches did you consider, and why did you settle on this one?
+
+#### Prompt remains a request
```
benbrandt: I am kind of torn at the moment of which method I prefer here...

Most of the agent notifications are still needed, it is just a matter of if the client emits session/updates or continues to call methods, but the response lifecycle is different.

docs/rfds/v2-prompt.mdx
```diff
@@ -0,0 +1,306 @@
...
+
+#### User message accepted/acknowledged
+
+In order to have a consistent understanding between agent and client on where the user message appears within the session history in relation to other messages, it is important to see when and where the agent has accepted the user message into the feed.
+
+This will also be important for queueing messages, depending on how we implement that, so that the client can know if it is still allowed to edit the queued message, or where in the turn order it got inserted.
+
+Even without a new queue, which may allow for editing the queued message, it means that the client doesn't necessarily have to send a `session/cancel` before prompting. This would need some exploration, but potentially the agent could decide whether it cancels the current turn and inserts it immediately, or inserts it at the next convenient break point. This should probably still be defined as "as soon as possible" and queueing would enable some later points, but it could still be more graceful than needing to cancel all current tool calls for example, as is required at the moment.
```
CSRessel, line 63: Definitely interested in what this enables for "steering" messages! That don't require waiting for a full turn completion to be accepted by the agent

docs/rfds/v2-prompt.mdx
```diff
@@ -0,0 +1,306 @@
...
+
+### Additional Agent `session/update` notification types
+
+As far as I can tell at the moment, we only need to add a few notification types on the agent side to make this work.
+
+#### User message accepted/acknowledged
+
+In order to have a consistent understanding between agent and client on where the user message appears within the session history in relation to other messages, it is important to see when and where the agent has accepted the user message into the feed.
+
+This will also be important for queueing messages, depending on how we implement that, so that the client can know if it is still allowed to edit the queued message, or where in the turn order it got inserted.
```
CSRessel, line 61: This is the one piece that gives me a little bit of pause

Before this, the structure of ACP messages had the nice effect of tracking exactly the back and forth of the agent and the user. It was a very clean separation of agent state, between runtime (agent) and interface (client). With this change, implementing a client feature like editing a queued message becomes a series of exchanges over the wire for each edit. I guess you can say the same for a config change like plan mode or model? But that state truly does live with the agent; this feels like deferring UI state into the backend

I don't think it's easy to implement something like steering messages without doing either 1. this, or 2. agent can ask to receive a queued message while still in the running state. The later seems uglier!

CSRessel:
I like what this aims to support. In part for multi-client replay, but particularly for agent initiated interactions outside of user prompts :+1:

docs/rfds/v2-prompt.mdx
chazcb: I tend to prefer that clients use methods for all critical comms so we can confirm the agent received it, but with a {} empty response. Notifications used for non-critical / idempotent client > agent data.

docs/rfds/v2-prompt.mdx
benbrandt: Agreed. As I wrote this out I became less and less enamored with the fully notification based approach.

I updated to my current thinking which I think is much nicer, and hopefully allows for better exploration of how to handle this.

docs/rfds/v2-prompt.mdx
benbrandt: Totally agreed. As I was writing this draft, I slowly became less excited about my own proposal 😆 

I updated with a my new thinking on this. Let me know what you think.

PR #986, opened 2026-04-15
RFD: session/status — session liveness check

chazcb:
## Summary

Proposes adding a `session/status` method — a side-effect-free, read-only probe that checks whether an agent is currently handling a specific session.

### Problem

ACP has no way to ask an agent "are you currently handling this session?" without side effects. Every existing session method mutates state:

- `session/load` forces the agent to start handling the session
- `session/prompt` sends work to the session
- `session/attach` attaches a client and sets up roles
- `session/resume` resumes and starts handling the session

If a client or routing layer wants to know "is this session live — will other session methods work?" before committing, there is no option.

### Proposal

```json
{ "method": "session/status", "params": { "sessionId": "sess_abc123" } }
→ { "result": { "status": "live" } }      // agent is handling this session
→ { "result": { "status": "not_found" } }  // agent is not handling this session
```

If `"live"`, the client can trust that `session/prompt`, `session/cancel`, and other session methods will work. If `"not_found"`, the client should use `session/load` to start the session.

**Contract:** must be side-effect-free, cheap, and safe to call frequently.

### Use cases

- **Client reconnect**: check if the session is still live before deciding between `session/prompt` (already live) and `session/load` (needs to be started)
- **Routing layer probe**: in multi-pod deployments, locate which pod is handling a session without forcing a load on the wrong pod
- **UI feedback**: show session availability immediately on reconnect


docs/rfds/session-status.mdx
```diff
@@ -0,0 +1,124 @@
...
+```
+
+Response when the session is not loaded:
+
+```json
+{
+  "jsonrpc": "2.0",
+  "id": 1,
+  "result": {
+    "status": "not_found"
```
benbrandt, line 69: Thoughts on returning the Resource not found error code?

docs/rfds/session-status.mdx
```diff
@@ -0,0 +1,124 @@
...
+```
+
+Response when the session is currently loaded:
+
+```json
+{
+  "jsonrpc": "2.0",
+  "id": 1,
+  "result": {
+    "status": "live"
```
benbrandt, line 57: I am not sure exactly how this would look, but I wonder about returning current position once we reliably have ids... like user message id, agent message id, tool call, etc?

Mostly helpful for whenever we allow for resuming at certain points, theoretical for now, I wouldn't add it to the RFD yet

docs/rfds/session-status.mdx
```diff
@@ -0,0 +1,124 @@
...
+  "jsonrpc": "2.0",
+  "id": 1,
+  "result": {
+    "status": "not_found"
+  }
+}
+```
+
+- `"live"` — The agent is currently handling this session. Other session methods (`session/prompt`, `session/cancel`, etc.) will work as expected.
+- `"not_found"` — The agent is not handling this session. It was never created here, has been closed, or was lost to a crash/restart. The client should use `session/load` to start handling it.
```
benbrandt, line 75: I wonder if there are 3 states?

1. Currently active/in-memory/running
2. able to be loaded/resumed
3. Never heard of it before

PR #1063, opened 2026-04-23
RFD: Sandbox Capability and Policy

visionik:
**Author:** [visionik](https://github.com/visionik) (OpenClaw maintainer)

## RFD file

[docs/rfds/sandbox-capability-policy.mdx](https://github.com/visionik/agent-client-protocol/blob/rfd/sandbox-capability-policy/docs/rfds/sandbox-capability-policy.mdx)

## Summary

Proposes two additive, capability-negotiated types and one relation:

- `SandboxCapability` — advertised by the agent in `AgentCapabilities.sandbox`; static per agent; describes the isolation the runtime actually enforces (mode + filesystem/network/process-capability guarantees).
- `SandboxPolicy` — set by the client per session via the existing `acp.sandbox` Session Config Option; describes the isolation the operator requires for this session's work.
- `satisfies(capability, policy)` — a deterministic predicate the agent evaluates at `setSessionConfigOption` time; rejects with a typed `sandbox_policy_unsatisfiable` error carrying the agent's capability if the policy cannot be met.

## Motivation

ACP has no first-class model for sandbox isolation today. Implementers are reinventing it in incompatible ways:

- **`_meta` knobs without a schema.** [agentpool's community reference](https://github.com/phil65/agentpool/blob/main/acp_meta_field_reference.md) documents `claude-code-acp` using `_meta` in `NewSessionRequest` for per-session configuration. Codex goes further: [PR #17763](https://github.com/openai/codex/pull/17763) adds a `codex/sandbox-state-meta` experimental capability that injects `SandboxState` into `_meta` on MCP tool calls so MCPs can opt in to receiving sandbox config. These are not discoverable via capability negotiation, not versioned, and collide in namespace across agents.
- **Silent host-only assumption.** Codex exposes `sandbox_mode` as a local `config.toml` key (`read-only` / `workspace-write` / `danger-full-access`) but nothing about the agent's sandbox posture is surfaced over ACP. A client routing work to a Codex ACP session has no protocol-level way to know what isolation it is actually getting.
- **Boolean collapse.** OpenClaw needed to enforce that sandbox-required child sessions could not be routed to ACP backends (which run on the host). Because there was no structured capability to check, the only safe option was a hard static rejection of `sandbox="require"` for `runtime="acp"` — documented in [PR #32254](https://github.com/openclaw/openclaw/pull/32254) and [GHSA-9q36-67vc-rrwg](https://github.com/openclaw/openclaw/security/advisories/GHSA-9q36-67vc-rrwg). A boolean `runsInSandbox` flag would have the same problem: it cannot distinguish Docker from chroot from seccomp, or filesystem isolation from network isolation.

The capability/policy split is the standard shape for this kind of check (LSP, MCP, WASI, Kubernetes RBAC) — this RFD proposes adopting it in ACP before the ecosystem fragments further.

## Relationship to in-flight work

The RFD includes a full accounting of adjacent in-flight RFDs and maintainer priorities that OpenClaw intends to adopt. This proposal covers only the one gap where no existing work was found.

## What's in the file

Full proposal including: elevator pitch, status quo with concrete failure modes, typed schema for both types, the `satisfies` predicate, rationale for Session Config Options placement, backward compatibility analysis, SDK rollout plan, open questions, FAQ (9 questions), and alternatives considered (5 alternatives with rejection rationale).

### DIFF
--- a/docs/rfds/sandbox-capability-policy.mdx
+++ b/docs/rfds/sandbox-capability-policy.mdx
@@ -1,0 +1,245 @@
+---
+title: "Sandbox Capability and Policy"
+---
+
+Author(s): [visionik](https://github.com/visionik) (OpenClaw)
+
+<Note>
+  **RFD draft status.** This file is authored in the OpenClaw repository as
+  `docs/refactor/acp-rfd-sandbox-capability-policy.mdx` for iteration and review with OpenClaw
+  maintainers and OpenClaw's upstream sponsors before being submitted to
+  `agentclientprotocol/agent-client-protocol` as `docs/rfds/sandbox-capability-policy.mdx` per the
+  [ACP RFD process](https://agentclientprotocol.com/rfds/about).
+</Note>
+
+## Relationship to in-flight ACP RFDs and maintainer work
+
+OpenClaw currently invokes agents through five different launch surfaces driven by four different execution engines. I (as an OpenClaw maintainer) am proposing that ACP could serve as a single consistent execution contract across all of them (see `docs/refactor/acp-everywhere.md`); to be clear, this proposal is in early discussion and may or may not be adopted by OpenClaw.
+
+While auditing what ACP already provides against what that consolidation would need, I found that the overwhelming majority of gaps are already covered by current RFDs and maintainer priorities. This RFD introduces the one area where I could find no existing work: typed sandbox modeling.
+
+Before the main proposal, a brief statement of how OpenClaw intends to consume adjacent in-flight work, so reviewers can see this RFD fits within the wider plan rather than competing with it:
+
+- **[Authentication Methods](https://agentclientprotocol.com/rfds/auth-methods)** — adopt directly; replaces OpenClaw's current ad-hoc auth-profile shuttling through `_meta`.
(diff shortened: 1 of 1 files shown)


anna239:
@visionik hi! Have you considered having a special config option in [Session Config Options](https://agentclientprotocol.com/protocol/session-config-options#session-config-options)? It could be a special flag, similar to how it's done for [model and thinking level](https://agentclientprotocol.com/protocol/session-config-options#option-categories)

visionik:
> @visionik hi! Have you considered having a special config option in [Session Config Options](https://agentclientprotocol.com/protocol/session-config-options#session-config-options)? It could be a special flag, similar to how it's done for [model and thinking level](https://agentclientprotocol.com/protocol/session-config-options#option-categories)

@anna239 Yes, this proposal uses session config for two of its three components:

SandboxPolicy — set by the client per session via the existing acp.sandbox Session Config Option; describes the isolation the operator requires for this session's work.

satisfies(capability, policy) — a deterministic predicate the agent evaluates at setSessionConfigOption time; rejects with a typed sandbox_policy_unsatisfiable error carrying the agent's capability if the policy cannot be met.

the third part (not session config) is a way for an ACP server to advertise what sandboxing options it has:

SandboxCapability — advertised by the agent in AgentCapabilities.sandbox; static per agent; describes the isolation the runtime actually enforces (mode + filesystem/network/process-capability guarantees).

or am I misunderstanding your question?

Thank yoU!

anna239:
@visionik my point is that Config options mechanism actually allows you to declare which options there are for a particular config. So I'm not quite sure how this setting is different from agent mode or f.i. thinking level

benbrandt:
@ignatov I think this fits well into our discussion of the "orchestrator" surface that doesn't fit neatly into the agent scope (requires too much implementation for each agent) and is better suited for an area of the protocol for something between the agent and client that sets up the agent process.

Going to close this one for now and revisit with that work.

PR #1064, opened 2026-04-23
transports: Revisions on streamable-http/ws transport request patterns

alexhancock:
Per [discussion](https://agentclientprotocol.zulipchat.com/#narrow/channel/590951-transports/topic/Updated.20Reference.20Implementation.20of.20RFD.20721.20Available/near/587693507) in the Transports WG Zulip here is a revision to the request patterns for consideration.

# Summary

* A single `Acp-Connection-Id` scoped GET stream delivers all server -> client responses and messages within a connection
* POSTS now return immediately with 202 Accepted
* We now will require HTTP/2 for multiplexing and using a single TCP connection
* Websocket implementation is unchanged

cc @anna239 @benbrandt

### DIFF
--- a/docs/rfds/streamable-http-websocket-transport.mdx
+++ b/docs/rfds/streamable-http-websocket-transport.mdx
@@ -297,20 +297,20 @@ The agent task is spawned once per connection. A single GET SSE stream delivers 
 
 ### Comparing to MCP Streamable HTTP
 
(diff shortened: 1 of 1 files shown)


benbrandt (approved):

alexhancock:
(@benbrandt just pushed a one line formatting thing after your approve)

PR #1237, opened 2026-05-18
RFD: Client-Provided System Prompt

wpfleger96:
Proposes a standard way for clients to provide system-level instructions at session creation, so agents can deliver them to the LLM's system prompt slot.

ACP is currently the only major agent protocol without a dedicated system prompt mechanism. MCP, OpenAI, and Anthropic's Messages API all have one. Every ACP implementation has independently implemented non-standard workarounds: `_meta` extensions, user-message injection, filesystem indirection. This RFD proposes a single interoperable solution:

- A `systemPrompt` capability under `sessionCapabilities`, advertised at `initialize`, so clients have reliable feature detection before sending the field
- An optional `systemPrompt` field on `session/new`, typed as `ContentBlock[]` for consistency with `session/prompt`, with text as the universal baseline
- Two delivery modes via an optional `systemPromptMode` field: `append` (default and universal baseline) and `override` (double opt-in, gated behind an `override` sub-capability)
- Session-scoped and immutable; restored on `session/load`/`session/resume` and inherited by `session/fork`

Motivated by [Discussion #414](https://github.com/orgs/agentclientprotocol/discussions/414).

### DIFF
--- a/docs/rfds/client-system-prompt.mdx
+++ b/docs/rfds/client-system-prompt.mdx
@@ -1,0 +1,149 @@
+---
+title: "Client-Provided System Prompt"
+---
+
+Author(s): [@wpfleger96](https://github.com/wpfleger96) (Block — contributor to [Sprout](https://github.com/block/sprout) and [Goose](https://github.com/aaif-goose/goose))
+
+## Elevator pitch
+
+> What are you proposing to change?
+
+Allow clients to provide a system prompt when creating or prompting a session. Today, ACP has no mechanism for a client to deliver system-level instructions to an agent — the only content a client can send is user-role messages via `session/prompt`. This means behavioral instructions, persona definitions, and platform context all compete with user content in the same message role, never reaching the LLM's architecturally privileged system prompt slot.
+
+This proposal comes from building two ACP clients at Block — [Sprout](https://github.com/block/sprout), a Nostr-based team collaboration and messaging platform with managed AI agents, and [Goose](https://github.com/aaif-goose/goose), an open-source AI agent. Both have independently hit this gap and implemented non-standard workarounds. We'd like to solve it at the protocol level.
+
+We propose adding an optional `systemPrompt` string field to ACP, giving clients a standard way to provide system-level instructions that agents can deliver to the LLM's system prompt.
+
+## Status quo
+
+> How do things work today and what problems does this cause? Why would we change things?
+
+ACP's `session/new` accepts only `cwd` and `mcpServers`. `session/prompt` accepts only `sessionId` and `prompt` (a `ContentBlock[]` of user content). There is no `systemPrompt`, `instructions`, or equivalent field on any ACP method — confirmed across all 80+ type definitions in the v0.13.0 schema.
+
+This creates several problems for clients that need to provide behavioral instructions to agents:
+
+**Workarounds are fragmented and non-standard.** Every implementation works around the gap differently:
+- [`claude-agent-acp`](https://github.com/agentclientprotocol/claude-agent-acp) added a `_meta.systemPrompt` extension ([issue #90](https://github.com/agentclientprotocol/claude-agent-acp/issues/90), [PR #91](https://github.com/agentclientprotocol/claude-agent-acp/pull/91)) — but the spec states implementations "MUST NOT make assumptions about values at these keys," making `_meta` unsuitable for instruction delivery that agents are expected to act on. The implementation silently dropped `excludeDynamicSections` for months ([issue #581](https://github.com/agentclientprotocol/claude-agent-acp/issues/581)), demonstrating the fragility of non-standard extensions.
+- [Sprout](https://github.com/block/sprout) is a Nostr-based team collaboration and messaging platform where AI agents are first-class workspace members. Each managed agent has a persona (behavioral instructions, role definition, platform context) that the ACP harness needs to deliver as system-level instructions. Today, Sprout injects this as `[System]\n{content}` prepended to the user message on every turn — content that never reaches the LLM's system slot. Sprout's own [Persona Pack Spec](https://github.com/block/sprout/blob/main/crates/sprout-persona/PERSONA_PACK_SPEC.md) lists "true system prompt injection via ACP" as a planned feature, explicitly acknowledging the current approach is a workaround.
+- [Goose](https://github.com/aaif-goose/goose) builds its system prompt internally via `PromptManager` from config files. ACP clients cannot inject or override it through the protocol. Goose has [an open issue](https://github.com/aaif-goose/goose/issues/7596) documenting the same gap in terms of its "recipe" system — recipes are Goose's mechanism for injecting behavioral instructions into a session, and the issue notes "there is no recipe field on `NewSessionRequest` and no logic to load/apply a recipe to the new session's agent." Goose's `_goose/session/update_project` method works around this by sending a project ID that the agent resolves to a file on disk — a Goose-specific indirection, not a protocol-level solution.
+- Codex receives no system prompt from clients. The harness controls only `approval_policy` and `sandbox_mode`.
+
+**Instruction following degrades when instructions are in user messages.** OpenAI's Instruction Hierarchy paper ([arXiv 2404.13208](https://arxiv.org/abs/2404.13208)) found 63% improvement in prompt injection defense when models treat system prompts as higher-priority than user messages. Research on system message compliance ([SysBench, arXiv 2408.10943](https://arxiv.org/abs/2408.10943)) further demonstrates that LLMs are specifically trained and evaluated on their ability to follow system-role instructions — a capability that ACP clients currently cannot leverage. The system role is architecturally privileged in modern LLMs, not just conventional.
+
+**ACP is an outlier.** Every comparable protocol provides system prompt support:
+
+| Protocol | Field | Type | Placement |
+|----------|-------|------|-----------|
+| MCP `sampling/createMessage` | `systemPrompt` | `string` | Per-request |
+| OpenAI Assistants API | `instructions` | `string` | Creation + per-run override |
+| OpenAI Responses API | `instructions` | `string` | Per-request |
+| Anthropic Messages API | `system` | `string \| ContentBlock[]` | Per-request |
+
+Agent frameworks follow the same pattern: AutoGen has `system_message`, CrewAI has `role`/`goal`/`backstory`, LangGraph has `prompt`. Only Google's A2A protocol is similarly silent — but A2A is agent-to-agent transport, a fundamentally different abstraction layer than ACP's client-to-agent communication.
+
+## What we propose to do about it
+
+> What are you proposing to improve the situation?
+
+Add an optional `systemPrompt` string field to `session/new`, allowing clients to set behavioral instructions when creating a session.
+
+```json
+{
+  "jsonrpc": "2.0",
+  "id": 1,
+  "method": "session/new",
+  "params": {
+    "cwd": "/home/user/project",
+    "mcpServers": [],
+    "systemPrompt": "You are a code review specialist. Focus on security vulnerabilities, performance issues, and adherence to the team's style guide. Be thorough but concise."
+  }
+}
+```
+
+`session/new` is the natural place for this — it already serves as the configuration point for `cwd` and `mcpServers`, and system prompts are fundamentally session-scoped (they define the agent's behavioral context for the entire interaction). One canonical place to set instructions, fully backward compatible, minimal protocol surface.
+
+The system prompt is immutable after session creation. If a client needs to change the system prompt mid-session (e.g., a user updates agent persona settings during an active conversation), it must start a new session. Mid-session updates could be addressed in the future via a dedicated method (e.g., `session/update_system_prompt`) without conflicting with this proposal.
+
+### Design choices
+
+- **Type: `string`.** Plain string covers the vast majority of use cases and matches the cross-protocol consensus (MCP, OpenAI, AutoGen all use simple strings). Structured types (e.g., `ContentBlock[]`) can be considered in a future extension if multimodal system prompts become a real need.
+- **Optional field.** Agents that don't support client-provided system prompts continue working — they simply ignore the field. Agents MUST have their own default behavior and MUST NOT require a client system prompt to function.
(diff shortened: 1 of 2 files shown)


docs/rfds/client-system-prompt.mdx
```diff
@@ -0,0 +1,149 @@
...
+  systemPrompt?: string;  // NEW
+}
+```
+
+This is additive — no existing fields change, no breaking modifications.
+
+### Agent behavior
+
+When an agent receives a `systemPrompt`:
+- It SHOULD use the content as the LLM's system prompt (or incorporate it into its system prompt construction)
```
andymac4182: Should this be clear that its always additional to existing system prompt so its a clear set of additional instructions vs replacing the System Prompt.

wpfleger96: great catch, I think you're right this needs to be cleaner. I just pushed a change to the RFD to clarify that `systemPrompt` uses **additive semantics** - the client system prompt is incorporated alongside the agent's own built-in instructions, it doesn't replace them. Agents always retain their own identity, safety guardrails, and internal configuration.

the updated Design Choices and Agent Behavior sections should make this explicit now. thanks for flagging it!

alexhancock:
I think it's good

One bit of feedback that comes to mind: if it is to remain a "SHOULD" for agents to thread it to the llm's system prompt slot, it may be good for the client to know if the agent is ignoring it.

Two solutions:

1. Consider making it a MUST for the agent to attempt to handle it
2. Make some kind of notification or value that comes back on the `session/new` response indicating to the client whether it is being used with the llm

Number 2 may also be needed for error cases where the system prompt couldn't be inserted into a facility for this on the llm side.

wpfleger96:
@alexhancock thanks for the feedback! I agree with both and just pushed an update with these changes:

- upgrade from `SHOULD` to `MUST` for incorporating the system prompt. I couldn't find any real-world use cases where an ACP agent couldn't/wouldn't honor a system prompt so giving clients a stronger guarantee with `MUST` feels right here
- add `systemPromptStatus` field on the `session/new` response: it returns `accepted` or `rejected` (with a reason) so clients know whether their system prompt addition was used or not. This covers the error cases you mentioned and if the client doesn't send a `systemPrompt`, the field is omitted from the response

aleclarson:
> 2\. Make some kind of notification or value that comes back on the `session/new` response indicating to the client whether it is being used with the llm
> 
> Number 2 may also be needed for error cases where the system prompt couldn't be inserted into a facility for this on the llm side.

> * add `systemPromptStatus` field on the `session/new` response: it returns `accepted` or `rejected` (with a reason) so clients know whether their system prompt addition was used or not. This covers the error cases you mentioned and if the client doesn't send a `systemPrompt`, the field is omitted from the response

It seems wrong to have the session created if the `systemPrompt` option can't be applied. Wouldn't it be better for the agent to respond with [an error result](https://agentclientprotocol.com/protocol/overview#error-handling) instead?

wpfleger96:
> > 2. Make some kind of notification or value that comes back on the `session/new` response indicating to the client whether it is being used with the llm
> > Number 2 may also be needed for error cases where the system prompt couldn't be inserted into a facility for this on the llm side.
> 
> > * add `systemPromptStatus` field on the `session/new` response: it returns `accepted` or `rejected` (with a reason) so clients know whether their system prompt addition was used or not. This covers the error cases you mentioned and if the client doesn't send a `systemPrompt`, the field is omitted from the response
> 
> It seems wrong to have the session created if the `systemPrompt` option can't be applied. Wouldn't it be better for the agent to respond with [an error result](https://agentclientprotocol.com/protocol/overview#error-handling) instead?

@aleclarson good call, that's a much cleaner design. I just pushed an update that drops `systemPromptStatus` entirely so if the agent can't apply the system prompt, `session/new` returns a standard JSON-RPC error and the session isn't created. this avoids an ambiguous middle state where a session exists but without the client's intended instructions. let me know what you think!

wpfleger96:
We've built a working implementation of this RFD in [block/sprout#981](https://github.com/block/sprout/pull/981) — a Rust ACP agent (`buzz-agent`) that handles the proposed `systemPrompt` field in `session/new`.

Key implementation details:
- Layering semantics: `base_prompt.md` is always the foundation for all agents — it is never removed or replaced. The client-provided `systemPrompt` (agent persona) layers on top of the base prompt. Hints are appended after that. The final prompt structure is: `[base_prompt] + [systemPrompt persona] + [hints]`.
- Empty string treated as absent (no-op)
- Agent returns a JSON-RPC error if the combined prompt exceeds its internal size limit (512KB) — exercising the spec's "MUST return a standard JSON-RPC error" clause
- No capability advertisement — feature detection via protocol version as proposed
- Backward-compatible: agents that don't recognize the field ignore it per standard JSON-RPC

docs/rfds/client-system-prompt.mdx
```diff
@@ -0,0 +1,149 @@
...
+}
+```
+
+`session/new` is the natural place for this — it already serves as the configuration point for `cwd` and `mcpServers`, and system prompts are fundamentally session-scoped (they define the agent's behavioral context for the entire interaction). One canonical place to set instructions, fully backward compatible, minimal protocol surface.
+
+The system prompt is immutable after session creation. If a client needs to change the system prompt mid-session (e.g., a user updates agent persona settings during an active conversation), it must start a new session. Mid-session updates could be addressed in the future via a dedicated method (e.g., `session/update_system_prompt`) without conflicting with this proposal.
+
+### Design choices
+
+- **Type: `string`.** Plain string covers the vast majority of use cases and matches the cross-protocol consensus (MCP, OpenAI, AutoGen all use simple strings). Structured types (e.g., `ContentBlock[]`) can be considered in a future extension if multimodal system prompts become a real need.
```
benbrandt: there is an argument to be had to keep this consistent.
I wonder if we could express this via capabilities which types it supports, but still use the standard content types

docs/rfds/client-system-prompt.mdx
```diff
@@ -0,0 +1,149 @@
...
+  systemPrompt?: string;  // NEW
+}
+```
+
+This is additive — no existing fields change, no breaking modifications.
+
+### Agent behavior
+
+When an agent receives a `systemPrompt`:
+- It MUST incorporate the content into its system prompt construction alongside its own built-in instructions
```
benbrandt: We need to have the agent express this as a capability or an "optional" field becomes a huge breaking change.

So, I think the agent needs to have a session capability of whether it supports system prompts, maybe also which modalities it supports, similar to the prompt capabilities.

I think we need this not only to express these MUSTs, but also for the client to know if it will be respected at all.

docs/rfds/client-system-prompt.mdx
```diff
@@ -0,0 +1,149 @@
...
+- It MUST NOT discard its own safety guardrails, identity, or internal configuration in favor of client-provided content
+- It MUST NOT require a client system prompt to function — agents always have a default
+- If the agent cannot incorporate the system prompt, it MUST return a standard JSON-RPC error and MUST NOT create the session
```
benbrandt: Again, all existing agents woulnd't know about this field, and therefore would be in violation of all of these.
We need to let the agents opt-in, and then we can hold them to this behavior

docs/rfds/client-system-prompt.mdx
```diff
@@ -0,0 +1,149 @@
...
+}
+```
+
+This is additive — no existing fields change, no breaking modifications.
+
+### Agent behavior
+
+When an agent receives a `systemPrompt`:
+- It MUST incorporate the content into its system prompt construction alongside its own built-in instructions
+- It MUST NOT discard its own safety guardrails, identity, or internal configuration in favor of client-provided content
```
benbrandt: I think we need to be clear on what sort of system prompt edits are allowed.
I believe claude allows you to append or override. 
We need to decide if we support both or not, or if we only start with append, we may still want to provide more structure and a type field so that we could change it in the future.

Not a hard requirement, but something to think about

docs/rfds/client-system-prompt.mdx
```diff
@@ -0,0 +1,149 @@
...
+- It MUST NOT require a client system prompt to function — agents always have a default
+- If the agent cannot incorporate the system prompt, it MUST return a standard JSON-RPC error and MUST NOT create the session
+
+When `systemPrompt` is absent:
+- The agent uses its own default system prompt (current behavior, unchanged)
+
+### Backward compatibility
+
+The field is optional with no default behavior change. Existing clients that don't send `systemPrompt` see no difference. Agents implementing this spec version MUST handle `systemPrompt` when provided — either incorporating it successfully or returning an error. Agents on older spec versions that don't recognize the field will ignore it per standard JSON-RPC behavior. Clients can determine whether an agent supports `systemPrompt` based on the protocol version negotiated during `initialize`.
```
benbrandt: protocolVersion is only major version. So unless we make this a hard requirement I don't see this being enough of a symbol.

It is also not clear to me that every agent would support this. I think it is highly likely that MOST will. But it still feels in the realm of a capability to me. Mainly because, if we require this field, but the agents can't express they don't support it, clients will just get lots of errors and have to remember _not_ to pass it in the next time.

docs/rfds/client-system-prompt.mdx
```diff
@@ -0,0 +1,149 @@
...
+
+The spec should clarify that these are independent mechanisms. A reasonable default: the client system prompt takes precedence for LLM instruction content, while config options continue to control tool availability, permissions, and other agent-internal configuration.
+
+### Should agents advertise system prompt support via capabilities?
+
+No separate capability flag is needed. The `session/new` response itself serves as acknowledgment: if the call succeeds, the system prompt was incorporated. If the agent cannot honor it, `session/new` returns a standard JSON-RPC error and no session is created — the client gets an explicit failure rather than a silently degraded session. For agents on older spec versions that don't recognize the field, clients can check the protocol version negotiated during `initialize` to determine support.
```
benbrandt: I've hammered this home enough I think, but an old agent won't error AND will silently ignore it.

benbrandt (changes requested):

docs/rfds/client-system-prompt.mdx
wpfleger96: agreed, a capability is the right mechanism here. I've added a `systemPrompt` capability under `sessionCapabilities` mirroring the existing pattern: omitted/null means unsupported, `{}` means supported. that also lets me scope the MUSTs to agents that actually advertise it. on modalities, I went with text as the universal baseline and left richer types as future sub-capabilities (more on that in the line 69 thread).

docs/rfds/client-system-prompt.mdx
wpfleger96: good point, I updated this so that all the agent-side MUSTs are now scoped to agents that advertise the new capability, so it's opt-in first and then enforceable.

docs/rfds/client-system-prompt.mdx
wpfleger96: great point, protocolVersion being major-only makes it useless for detecting an additive field like this. I dropped that claim entirely. the capability from your other comment is now the detection mechanism, and clients MUST NOT send the field to agents that don't advertise it, so nobody ends up in the error-and-remember loop you described.

docs/rfds/client-system-prompt.mdx
wpfleger96: rewrote it to a "yes" with the capability design, and the backward-compat section now explains why the hazard is closed by construction (same approach the additional-directories RFD took).

docs/rfds/client-system-prompt.mdx
wpfleger96: makes sense. once the capability object exists anyway, using `ContentBlock[]` is free and keeps the protocol consistent, so I switched the type. one deliberate divergence from the `session/prompt` baseline: I chose **text-only** as the required baseline (not text + resource_link). the reasoning is that the other variants don't translate to the system slot the way they do to a prompt turn. `resource_link` would force the agent to fetch and resolve the URI before session creation with unspecified failure semantics (missing/huge/unreachable resources, snapshot vs re-read), and `image`/`audio` can't land in the system slot at all since it's text-only in every major LLM API (Anthropic's `system`, OpenAI's `instructions`, Gemini's `systemInstruction`).

docs/rfds/client-system-prompt.mdx
wpfleger96: the RFD now specs both modes: append is the default and universal baseline (every advertising agent MUST support it, agent keeps its own guardrails), and override is double opt-in. the agent advertises an `override` sub-capability (`systemPrompt: { "override": {} }`) and the client explicitly requests it via an optional `systemPromptMode` field on `session/new` (defaults to append when omitted). I went with a sibling mode field instead of an envelope object so the common append case stays a bare content array. that gives us your future-proofing structure, just capability-gated so there's no silent mode mismatch.

docs/rfds/client-system-prompt.mdx
```diff
@@ -0,0 +1,262 @@
...
+
+This proposal comes from building two ACP clients at Block — [Buzz](https://github.com/block/buzz), a Nostr-based team collaboration and messaging platform with managed AI agents, and [Goose](https://github.com/aaif-goose/goose), an open-source AI agent. Both have independently hit this gap and implemented non-standard workarounds. We'd like to solve it at the protocol level.
+
+We propose:
+
+- A `systemPrompt` capability under `sessionCapabilities`, advertised by agents at `initialize`, so clients know before creating a session whether system prompt delivery is supported.
+- An optional `systemPrompt` field on `session/new`, typed as `ContentBlock[]` for consistency with `session/prompt`, with text content as the universal baseline.
+- Two delivery modes: **append** (the default and universal baseline — client instructions layer on top of the agent's own system prompt) and **override** (client content replaces the agent's base prompt), the latter gated behind an `override` sub-capability.
```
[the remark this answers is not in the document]
wpfleger96: fixed, PR description rewritten to match the current proposal (capability + `ContentBlock[]` + append/override modes).

docs/rfds/client-system-prompt.mdx
```diff
@@ -0,0 +1,262 @@
...
+
+## Status quo
+
+> How do things work today and what problems does this cause? Why would we change things?
+
+ACP's `session/new` accepts only `cwd`, `mcpServers`, and `additionalDirectories`. `session/prompt` accepts only `sessionId` and `prompt` (a `ContentBlock[]` of user content). There is no `systemPrompt`, `instructions`, or equivalent field on any ACP method — confirmed across the full v1 schema (v1.20.0).
```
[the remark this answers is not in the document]
wpfleger96: fair point, updated the wording to carve out `_meta` explicitly.

docs/rfds/client-system-prompt.mdx
```diff
@@ -0,0 +1,262 @@
+---
+title: "Client-Provided System Prompt"
+---
+
+Author(s): [@wpfleger96](https://github.com/wpfleger96) (Block — contributor to [Buzz](https://github.com/block/buzz) and [Goose](https://github.com/aaif-goose/goose))
```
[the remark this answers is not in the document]
wpfleger96: good catch, added `rfds/client-system-prompt` to the Draft group in `docs/docs.json`.

PR #1214, opened 2026-05-16
docs(rfd): Session Rewind

htahaozlu:
This RFD proposes adding `session/rewind` and `session/edit_prompt` methods to ACP, enabling user-facing conversation rewind and message editing — features universally present in coding-agent UIs (Claude Desktop, Cursor, Windsurf, Gemini CLI checkpointing) but currently lacking a portable protocol surface.

Builds on `unstable_message_id` (already in `Cargo.toml`'s `unstable` set).

Complementary to, and distinct from, [`session/fork`](https://agentclientprotocol.com/rfds/session-fork): fork is non-destructive context copying for side queries; rewind mutates the session timeline in place.

Consolidates two previously scattered discussions:

- [#239](https://github.com/orgs/agentclientprotocol/discussions/239) — Restoring a checkpoint and an earlier point in the conversation
- [#329](https://github.com/orgs/agentclientprotocol/discussions/329) — `session/undo` and `session/redo` requests

Filesystem rollback is explicitly out of scope (see Status quo and FAQ in the RFD body) — left to existing `fs/*` methods, `session/fork`, IDE-native snapshots, or the user's SCM.

Concrete user-facing bugs in Zed that motivate a portable protocol surface:

- [zed#52153](https://github.com/zed-industries/zed/issues/52153) — ACP with Codex CLI hangs: "failed to get old checkpoint — not implemented yet"
- [zed#39997](https://github.com/zed-industries/zed/issues/39997) — Zed removes all AI changes when editing a message with no checkpoints consideration
- [zed#28676](https://github.com/zed-industries/zed/issues/28676) — Agent Panel: Undo/Redo around checkpoints, checkpoint restore irreversibly destructive
- [zed#55888](https://github.com/zed-industries/zed/issues/55888) — Claude Code ACP: conversation after `/clear` is lost on Zed restart

**Status**: Draft proposal, open for champion from the core team. Happy to iterate on the design — open to splitting into separate RFDs (rewind vs edit-prompt), merging with related work, or scoping down to a single primitive if preferred.

CLA: will sign at https://zed.dev/cla before any implementation PR. Posting this RFD PR per the [RFD process](https://agentclientprotocol.com/rfds/about) so the design discussion can happen against concrete text.

PR #1261, opened 2026-05-19
docs(rfd): mid-turn input via session/inject (queue and steer)

kennethsinder:
Lifts [discussion #1220 (`session/inject` for mid-turn queue + steer)](https://github.com/orgs/agentclientprotocol/discussions/1220) into the formal RFD process, per @benbrandt's offer to put it in the v2 bucket.

## Summary

One method (`session/inject`), two modes (`queue`, `steer`), one capability. Built on the [v2 prompt lifecycle](https://github.com/agentclientprotocol/agent-client-protocol/blob/main/docs/rfds/v2/prompt.mdx): agent-owned `messageId`, `user_message` echo notification, and `state_change`. Specifies the protocol shape for queue/steer behavior already present across Cursor, Codex CLI, Claude Code, Windsurf Cascade, Gemini CLI, and others.

## Relation to existing work

- **Supersedes [PR #484](https://github.com/agentclientprotocol/agent-client-protocol/pull/484)** (prompt queueing capability) with credit to @SteffenDE. That PR predates the v2 prompt lifecycle and doesn't address steer, delivery, or pending-message edit semantics. I can coordinate on close/redirect if the maintainers prefer.
- **Composes with** [message-id](https://agentclientprotocol.com/rfds/message-id), [request-cancellation](https://agentclientprotocol.com/rfds/request-cancellation) for in-flight request cancellation before an inject is accepted, [multi-client attach #533](https://github.com/agentclientprotocol/agent-client-protocol/pull/533), [session rewind #1214](https://github.com/agentclientprotocol/agent-client-protocol/pull/1214), [turn-complete signal #644](https://github.com/agentclientprotocol/agent-client-protocol/pull/644).
- **Sibling but separate from** [discussion #1224 `session/remind`](https://github.com/orgs/agentclientprotocol/discussions/1224) (system-role context injection; same break-point machinery, different payload contract).

## Looking for

A champion. @benbrandt offered in the discussion thread; tagging here. I can iterate on framing, split scope, or rework around the v2 prompt lifecycle if a different factoring lands better.

## Changes in this PR vs. the discussion post

- Tightened `messageId` ownership to agent-owned, returned in the inject response, matching the v2 prompt lifecycle's resolution.
- Added explicit acknowledgment and supersession of PR #484.
- Replaced `$/cancel_request` revocation with `messageId`-based pending operations: `session/revoke_inject`, optional `session/replace_inject`, and defined sad paths for already-delivered and unknown messages.
- Took positions on pending behavior: pending injects survive `session/cancel` by default; steer is held, not dropped, during `session/request_permission`.
- Expanded prior art with Windsurf Cascade, Gemini CLI (including its own [open `/inject` proposal](https://github.com/google-gemini/gemini-cli/issues/17197)), and the explicit cross-ecosystem "disambiguate queue/steer" discussions.

---

Process note: drafting used AI for outline and prior-art search; cited links were checked.


docs/rfds/v2/session-inject.mdx
```diff
@@ -0,0 +1,318 @@
...
+
+PR [#484](https://github.com/agentclientprotocol/agent-client-protocol/pull/484) (open since February 2026) takes a thinner cut: a `promptQueueing` capability plus an `end_turn` early-finish signal, with prompts queued via a parallel `session/prompt` call. It doesn't address steer, doesn't define delivery semantics, and predates the v2 prompt lifecycle's `state_change` and `user_message` notifications. The follow-up thread also asks how clients edit a queued message. This RFD answers that directly: editing is a pending-inject operation before `user_message`, not a transcript rewrite after delivery. The author also flags that the Claude Agent SDK had no public hook for "when was my queued message inserted," which is precisely what `user_message` echoing solves at the protocol level. The proposal below is intended to supersede #484, with credit to [@SteffenDE](https://github.com/SteffenDE) for raising the question first.
```
SteffenDE: > It doesn't address steer, doesn't define delivery semantics

I'd say this is not a correct representation of that PR.

promptQueueing was always meant to be steering. If you just want to queue for sending after the current turn, this is trivial to implement by ACP clients.

The early end_turn was the signal to the client when the new message was injected into the context.

docs/rfds/v2/session-inject.mdx
```diff
@@ -0,0 +1,318 @@
...
+      {
+        "type": "text",
+        "text": "stop using the legacy auth path, see auth_v2.go",
+      },
+    ],
+  },
+}
+```
+
+The agent responds when the inject has been accepted for pending delivery, not when the model has processed it. The response carries the agent-assigned `messageId`, matching the v2 prompt lifecycle pattern:
```
SteffenDE, line 64: Just as a note: this won't work for the Claude Agent SDK, because it will only generate the id at the point of delivery. In practice, this means that ACP message IDs would be an adapter concern, requiring the ACP code to store its own state persistently somewhere in order to get the same IDs for later loads.

We already have similar problems for message IDs in streaming chunks, so at some point this is probably inevitable.

docs/rfds/v2/session-inject.mdx
```diff
@@ -0,0 +1,318 @@
...
+}
+```
+
+An agent that only supports buffering at end-of-turn declares `["queue"]`. A tool-loop agent that can break safely between tool calls declares `["queue", "steer"]`. A streaming-only agent that can't do either declares the absence of the capability and clients fall back to `session/cancel`.
+
+Revoke is part of the pending-inject contract. `pending.replace` is optional. Clients should only offer edit-in-place for already-sent pending messages when it is advertised; otherwise they can keep drafts client-side longer or revoke and send a new inject when losing queue position is acceptable.
+
+### Semantics
+
+**`queue`** is the simple one. The agent buffers the content. The agent delivers it as a `user_message` notification once `state_change: idle` fires for the current turn. FIFO across multiple queued injects. If a queue lands on an already-idle session, the agent treats it as a normal user message and starts a turn, though clients should prefer `session/prompt` in that case.
```
SteffenDE: Maybe worth discussing if this needs to be part of ACP in the first place, as clients need to implement logic for displaying queued messages anyway, so adding the logic to buffer those client-side and do regular prompt on end_turn might be enough and keeps the protocol smaller. No strong opinions on my side there though.

docs/rfds/v2/session-inject.mdx
```diff
@@ -0,0 +1,318 @@
...
+        "pending": { "replace": true },
+      },
+    },
+  },
+}
+```
+
+An agent that only supports buffering at end-of-turn declares `["queue"]`. A tool-loop agent that can break safely between tool calls declares `["queue", "steer"]`. A streaming-only agent that can't do either declares the absence of the capability and clients fall back to `session/cancel`.
+
+Revoke is part of the pending-inject contract. `pending.replace` is optional. Clients should only offer edit-in-place for already-sent pending messages when it is advertised; otherwise they can keep drafts client-side longer or revoke and send a new inject when losing queue position is acceptable.
```
SteffenDE: This doesn't seem clear to be. Is revoke required or not? Can an agent omit `pending` and therefore signal that revoke is not supported?

docs/rfds/v2/session-inject.mdx
```diff
@@ -0,0 +1,318 @@
...
+- If delivery wins, the agent returns an error. The recommended shape is `-32602 Invalid params` with `error.data.reason: "already_delivered"` and the `messageId`. The client should keep the delivered message in the transcript.
+- If the `messageId` is unknown for that session, return `-32602 Invalid params` with `error.data.reason: "unknown_message_id"`.
+- If the message was already revoked and the agent still remembers the tombstone, return success. This makes revoke safe to retry after transport loss.
+
+Normal session errors still apply. If the session is unknown, closed, or no longer accepts input, return the same error the agent uses for other session requests.
+
+Delivery is the point where the agent commits the inject to session history and emits, or has irrevocably queued, the matching `user_message`. From that point, the content may already be in the next model input. Revoke must return `already_delivered`, and the client should expect the `user_message` if it has not seen it yet. There is no separate "revoke a delivered message" surface; that is [session rewind (#1214)](https://github.com/agentclientprotocol/agent-client-protocol/pull/1214) territory.
```
SteffenDE: > or has irrevocably queued

I'd say an agent should always emit the user_message at the point where it belongs in the chat. If a message is only "irrevocably queued" meaning that is can't be edited any more, but belongs to a later point in history, I'd say that should be an error on revoke / replace.

docs/rfds/v2/session-inject.mdx
```diff
@@ -0,0 +1,318 @@
...
+
+### Interaction with other in-flight requests
+
+- **`session/cancel` during a pending steer.** Codex [#22815](https://github.com/openai/codex/issues/22815) is a real bug surface here, and we should resolve it in the spec rather than leave it implicit. Recommended behavior: cancel applies to the in-flight turn only. Pending injects (both queue and steer) survive `session/cancel` and deliver as normal once the agent reaches idle. Clients that want to clear everything can call `session/revoke_inject` for pending injects first, then `session/cancel`. Agents should document if they deviate.
+- **`session/request_permission` blocking the turn.** While the agent is awaiting a permission decision, the turn is paused but not idle. Queue accumulates as normal. Steer is held until the permission resolves, then delivered at the next break-point (which may be immediately if the permission decision is the break-point). Agents that prefer to drop steers during permission waits may do so, capability-advertised.
```
SteffenDE: > Agents that prefer to drop steers during permission waits may do so, capability-advertised.

I'd drop that. If an Agent (SDK) can't handle this, the ACP adapter code should buffer rather than just dropping. But maybe I'm misunderstanding the point.

SteffenDE (commented):
Some comments from my side. Don't see me as authoritative here though. Happy to close the `promptQueueing` PR, as it makes more sense to wait for v2 to land.

docs/rfds/v2/session-inject.mdx
kennethsinder: Fair pushback, that was a sloppier read than I should have given #484. Rewrote the paragraph to call your early-`end_turn` what it is: a steer-via-yield design where the agent yields and the client's parallel `session/prompt` becomes the next turn. The honest delta vs. this RFD is that yield-and-re-prompt forces a turn boundary at every steer, so a tool call mid-stream surfaces as completing a turn instead of being absorbed, and the client drives the re-prompt rather than handing the agent a payload. Both reach steer; the trade-offs differ.

docs/rfds/v2/session-inject.mdx
kennethsinder: Yeah, this is the part where I think the adapter has to pay the cost and we should just say so out loud. Added an FAQ explicitly addressing it: the wire ID has to come back synchronously, so any adapter bridging to an SDK that doesn't expose a pre-delivery ID needs to mint a wire ID at accept-time and remember the mapping until the underlying SDK delivers. Same shape as the streaming-chunk situation `message-id` already accepts. The two alternatives (client-provided ID, deferred ID-on-delivery) are worse for different reasons that the FAQ spells out.

docs/rfds/v2/session-inject.mdx
kennethsinder: Good question and one I went back and forth on. Added a dedicated FAQ ("Does `queue` need to be in the protocol at all?") that tries to be honest about it: for one client driving one agent, you're right, client-side buffering covers the user-visible behavior and the protocol stays smaller. The cases I think justify the protocol-level queue are multi-client fan-out (with #533, the agent has to be the ordering authority because client-side queues can't agree on FIFO), `session/load` reconnect (a queue at the agent survives client restart), and headless agents serving multiple thin surfaces. Open to dropping it if the consensus is that those cases are too speculative for v2.

docs/rfds/v2/session-inject.mdx
kennethsinder: Good catch, the wording was vague. Tightened it: revoke is mandatory if you support `session/inject` at all. Any agent that can accept a pending inject can drop one before delivery; worst case is a tombstone flag plus skipping the `user_message` emit. Replace stays opt-in via `pending.replace` because content editing is genuinely harder than dropping (the content may already be partially serialized into a prompt envelope by the time the replace lands). The asymmetry is now spelled out instead of implied.

docs/rfds/v2/session-inject.mdx
kennethsinder: Agreed, that hedge created a state that doesn't actually exist on the wire. Tightened the definition: delivery is exactly the `user_message` emission. Before that, revoke succeeds. At or after that, revoke returns `already_delivered`. Any internal commit the agent does before emitting is purely for its own bookkeeping.

docs/rfds/v2/session-inject.mdx
kennethsinder: Fair, that carve-out was a cop-out and pushed work onto agents that should sit in the adapter. Removed it. The spec is now uniform: pending steer is held during a permission wait and delivered at the next break-point after the decision resolves. If the underlying runtime can't hold it, the ACP adapter buffers. No capability flag for this.

kennethsinder:
Thanks for the careful review, @SteffenDE! Replied to each thread inline. Quick summary of what changed in [48a629f](https://github.com/agentclientprotocol/agent-client-protocol/pull/1261/commits/48a629f) + [c3c9980](https://github.com/agentclientprotocol/agent-client-protocol/pull/1261/commits/c3c9980):

- Reframed the comparison with #484 to acknowledge its early `end_turn` as a legitimate steer-via-yield design, and articulated the actual trade-off (forced turn boundary at every steer vs. absorb-into-running-turn).
- Made revoke mandatory; replace stays opt-in via `pending.replace`. Asymmetry explained.
- Dropped the "irrevocably queued" hedge. Delivery is now defined strictly as `user_message` emission; no phantom intermediate state.
- Dropped the agent-may-drop-steers-during-permission carve-out. Adapter buffers when the underlying runtime can't.
- New FAQ on the adapter burden for `messageId` (the Claude Agent SDK case): the wire ID has to come back synchronously, adapter mints + maps, same shape `message-id` already accepts for streaming chunks.
- New FAQ on why `queue` is in the protocol rather than client-side only: multi-client FIFO ordering authority, replay/reconnect, headless agents serving multiple thin surfaces.
- Prose pass: tightened em-dashes to colons where natural, eased back on rule-of-three flourishes, removed "not X. It is Y" constructions to match the tone of the v2 prompt lifecycle RFD.

Open to more rounds. The biggest open design question I think is the queue itself: whether the multi-client / replay / headless cases justify protocol-level standardization at v2 stabilization time, or whether it's cleaner to ship steer alone and let queue land later once #533 firms up.

htahaozlu:
Thanks for the clean factoring here, @kennethsinder — the inject-vs-rewind split you drew in #1220 held up, so I leaned on it.

I've re-based the session-rewind proposal onto the v2 prompt lifecycle and filed it as the **"Truncate/Edit support"** item from the [v2 overview](https://github.com/agentclientprotocol/agent-client-protocol/blob/main/docs/rfds/v2/overview.mdx)'s "RFDs to be Written" list: **#1321** (`docs/rfds/v2/session-rewind.mdx`). It supersedes my earlier v1-substrate draft (#1214).

The two line up the way you framed them:

- **inject = additive** — adds a `user_message`, never touches prior history.
- **rewind = destructive** — `session/rewind` truncates the tail, `session/edit_prompt` rewinds + re-runs.

Both ride the same substrate (agent-owned `messageId`, `user_message` echo, `state_change`). Inject reuses `user_message`; rewind adds one symmetric notification, `history_truncated`, for the removal case so multi-client and `session/load` replay converge. Capabilities (`session.inject` vs `session.rewind`) and flags don't collide, and they compose — inject a correction mid-turn, or rewind to redo the turn.

@benbrandt — since the v2 overview already lists Truncate/Edit as a to-be-written RFD, happy to iterate framing in the rewind PR thread, and glad to align it with however the prompt-lifecycle `unstable`/`_meta` flagging shakes out.

kennethsinder:
Rebased this onto the current `main` and resolved the `docs/docs.json` v2 RFD nav conflict.

Validation on the rebased branch:

```bash
node -e 'JSON.parse(require("fs").readFileSync("docs/docs.json","utf8")); console.log("docs.json ok")'\nnpx prettier --check docs/docs.json docs/rfds/v2/session-inject.mdx\n```

xxchan:
Production feedback from [Raft](https://raft.build): this RFD is not a nice-to-have for us; the absence of steering is currently an adoption blocker for ACP.

Raft is a multi-agent collaboration platform. A core interaction is that a user can send a correction or additional context from another client/channel while an agent is still executing. Waiting until the current turn becomes idle is too late for a correction, while the only standard ACP fallback today, `session/cancel` followed by a new `session/prompt`, discards in-flight work and reports tool calls as cancelled. We need a capability-advertised way to deliver a user message at the next safe breakpoint without ending the running turn.

This is now blocking a concrete runtime integration. Grok's public programmatic surfaces are ACP over stdio and WebSocket. We tested the current Grok CLI (0.2.93): a concurrent `session/prompt` is queued until the active turn completes; `session/cancel` plus a new prompt only provides abort-and-redirect; and Grok's private TUI interjection path is not registered on either public ACP transport. As a result, we cannot offer the same live correction contract that we can offer through Codex app-server's `turn/steer`.

Because of this gap, we have stopped treating ACP as a sufficient canonical runtime abstraction for Raft, and we cannot ship a production-quality Grok runtime through ACP today without explicitly degrading steering to destructive cancellation. This is exactly the kind of fragmentation ACP should remove: the runtime supports interjection internally, but the shared protocol cannot express it.

Strong +1 from an implementer/operator for moving `session/inject` forward. Could a maintainer champion this RFD and clarify the v2 path/timeline? If protocol-level queue semantics are holding up consensus, our experience argues for landing a steer-only capability first (even as an unstable v2 extension) and allowing queue/revoke/replace semantics to follow. The minimum useful contract for us is steer acceptance, delivery at a safe breakpoint, capability negotiation, and an observable user-message/history event.

We are happy to provide our Grok/ACP probe results and implementation feedback against a narrowed proposal.

ChrisAkre:
I found this proposal yesterday and I implemented a copilot sdk -> acp bridge that added `session/inject` capability and it is an utterly massive quality of life improvement.

On the slim chance this RFD is rejected, there is zero chance I'm ever going back to an ACP protocol without it. 

(note my usecase is a ACP<->websocket bridge that supports multi-client fanout so you can control multiple harnesses on multiple devices concurrently)

kennethsinder:
^ That is so cool, thanks everyone for sharing the love and helping shape this <3 cc @benbrandt let me know if I missed any comms from you or anything that I can provide to help smooth this proposal along, sorry for the radio silence the last few weeks!

danielkov:
I implemented the RFD’s `session/inject` in our ACP-based harness: Kit and plumbed it through the protocol schema, Rust SDK:

- Protocol schema: #2043
- JSON-RPC response ordering: agentclientprotocol/rust-sdk#338
- Rust SDK injection support: agentclientprotocol/rust-sdk#339
- `agentkit` implementation: danielkov/agentkit#13
- Kit integration: speakeasy-api/kit#24

The implementation currently supports `steer` with `finish` behavior and revoke. Accepted messages are delivered FIFO after the current model stream or tool call finishes, and before the next model request.

One issue I ran into was ensuring the successful inject response is sent before the corresponding `user_message` delivery notification. In batched JSON-RPC responses, the full batch must be ready first. I added tracked response receipts in agentclientprotocol/rust-sdk#338 to provide that ordering.

Feedback appreciated!

PR #1269, opened 2026-05-21
docs: Add schema download note to schema page

benbrandt:


cmoulliard:
Links are not correct as they point here: https://github.com/agentclientprotocol/agent-client-protocol/releases/latest/download/schema.json even if this is for spec v1 or v2. 

Ideally it should point the tagged version of the spec file ;-)

benbrandt:
Understood. We don't currently publish v2, once we do I will update

PR #1281, opened 2026-05-22
feat(unstable): Add optional userCode to URL elicitation messages

anvilpete:
This allows the client UI to expose a code to the user for device code auth (with a click-to-copy button, for example).

### DIFF
--- a/schema/schema.v2.unstable.json
+++ b/schema/schema.v2.unstable.json
@@ -2713,6 +2713,10 @@
           "description": "The URL to direct the user to.",
           "format": "uri",
           "type": "string"
+        },
+        "userCode": {
+          "description": "If provided, the user must visit the URL and enter this code to complete the elicitation.",
+          "type": ["string", "null"]
         }
       },
       "required": ["elicitationId", "url"],
--- a/docs/protocol/draft/schema-v2.mdx
+++ b/docs/protocol/draft/schema-v2.mdx
@@ -1903,6 +1903,10 @@ URL-based elicitation where the client directs the user to a URL.
 <ResponseField name="url" type={"string"} required>
   The URL to direct the user to.
 </ResponseField>
+<ResponseField name="userCode" type={"string | null"}>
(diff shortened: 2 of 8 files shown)


schema/schema.v2.unstable.json
```diff
@@ -2713,6 +2713,10 @@
           "description": "The URL to direct the user to.",
           "format": "uri",
           "type": "string"
+        },
```
anna239, line 2716: should we maybe include some text description into the elicitation schema? Then agent can provide any instructions there

anvilpete: @anna239 do you mean in addition to [the existing `message` field](https://agentclientprotocol.com/protocol/v1/draft/schema#param-message) in the top-level elicitation request? If so, how would this new field differ?

anna239: oh, I've overlooked `message` for some reason, then the question is: why can't it be included in the `message`?

anvilpete: No worries!
It's useful to also have a field that semantically means "this is the code that the user needs to copy/paste", so the client can provide affordances to help (styling, click-to-copy button, etc.).

benbrandt: Thinking about this one some more... I am also wondering what is wrong with just using the message?
Currently agents are already doing something like this that use these in a generic string message with markdown formatting.

anvilpete: That's a good point, updating the `message` docs to make it clear that it can/should be markdown-formatted would help. I had assumed it was a plain text field.

But I think it's still useful for the client to know the `userCode`. A "click-to-copy" affordance needs this, for example, and at the moment this requires a regex on the `message`, which is far from ideal.

This is similar to how the `url` field provides semantic information to the client about the elicitation, rather than relying only on the opaque `message` string.

Still, if you both think this isn't useful, then I'm also happy to drop it.

benbrandt:
Let's close for now and see if this request resurfaces. thanks

PR #1292, opened 2026-05-27
docs(rfd): Expose client LSP servers over ACP

Ruddickmg:
Hello, first time contributing here. Let me know if there's anything else needed from me. Thanks!

### DIFF
--- a/docs/rfds/lsp-over-acp.mdx
+++ b/docs/rfds/lsp-over-acp.mdx
@@ -1,0 +1,80 @@
+---
+title: "LSP over ACP: Expose LSP servers over ACP"
+---
+
+<!--
+
+Instructions:
+
+* Copy this file and give it a name like my-feature-slug.mdx
+* Do not remove the `>` sections -- they are part of the template!
+* Replace the HTML comments that give general guidance with your answers and content. Do not begin your answers with `>`, your text should be unquoted.
+* For optional sections, you can leave the HTML comment as is if you do not wish to provide an answer.
+* In the FAQ, begin each question with a markdown `##` section so that it gets a linkable anchor.
+
+-->
+
+Author(s): [githubhandle](url)
+
+## Elevator pitch
+
+> Agents benefit from LSP access for code intelligence but are unable to use the client's already-configured LSP servers. Exposing the client's LSP servers over ACP would simplify configuration and enable agents to explore the code base with the same context as a user.
+
+<!--
+    Give a brief, high-level overview of what you plan to do and what problem you are solving. Feel free to use bullet points to help clarify the structure.
+-->
+
+## Status quo
+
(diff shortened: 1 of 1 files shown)


ignatov:
Hey! Thanks for your suggestion, the idea is clear, but I believe there're plenty of details we need to clarify before adopting that PR

Ruddickmg:
> Hey! Thanks for your suggestion, the idea is clear, but I believe there're plenty of details we need to clarify before adopting that PR

Of course! I mainly wanted to get the conversation started. Per the [RFD Process docs](https://agentclientprotocol.com/rfds/about#the-rfd-process), this is intended as a minimal initial proposal.

I avoided getting into implementation details since I’m not familiar with the codebases involved. At a high level, I was envisioning a lightweight bridge that would allow ACP clients to forward LSP requests between clients and agents.

If there's anything you'd like me to contribute—whether that's clarifying requirements, fleshing out the design, or helping with implementation—just let me know.

benbrandt:
I think the question at hand is:
- how many agents currently support LSPs (a few, but not a ton)
- How many would forward requests over ACP vs spawning it themselves (given how fs + terminal use via the client was adopted, I assume not much)

My hunch is most agents will want to own the LSP processes and communication. So then the problem becomes more of whether the agent would allow configuring LSPs from the session config or not...

Ruddickmg:
> I think the question at hand is:
> 
>     * how many agents currently support LSPs (a few, but not a ton)
> 
>     * How many would forward requests over ACP vs spawning it themselves (given how fs + terminal use via the client was adopted, I assume not much)
> 
> 
> My hunch is most agents will want to own the LSP processes and communication. So then the problem becomes more of whether the agent would allow configuring LSPs from the session config or not...

Most of the agents I’ve used (OpenCode, Copilot, Claude, etc.) support LSPs, so I had (perhaps incorrectly?) assumed this was a common capability.

To be clear, I’m not suggesting that ACP should facilitate configuring or managing a language server. I agree that it wouldn’t make much sense for a client to configure a language server that is otherwise controlled by an agent via ACP.

I’m suggesting that the agent communicate with an LSP that is already managed by the client (running in Neovim, VS Code, etc.). This would remove the need for the agent to configure or manage the language server at all. The agent would simply make LSP requests to the existing language server, with ACP acting as the bridge between the agent and the client-controlled LSP.

In effect, this would just expose a channel/endpoint through which the agent can communicate with the language server. From the agent’s perspective, it still communicates with the language server using LSP; the only difference is that, instead of connecting directly to an LSP that it manages, it communicates through the endpoint provided by ACP.

I think this could also make LSP integration simpler for agents. Rather than having each agent manage its own language servers or maintain integrations with specific language servers, the client could handle that entirely. The only requirement for the agent would be support for the Language Server Protocol, while ACP provides the connection between the agent and the client’s existing language server.

PR #1302, opened 2026-05-28
docs(rfd): Requested Tool Categories — optional ClientCapabilities hint for cross-vendor in-LLM tool exposure

s2agi:
## Summary

Opens an RFD proposing an optional `requested_tool_categories` field on `ClientCapabilities`, so a client can hint to the agent which **vendor-neutral categories** of in-LLM tools it would like exposed during a session.

This is purely the RFD document at `docs/rfds/requested-tool-categories.mdx`. No Rust changes yet — those would land in a follow-up PR if/when a core team member champions the proposal, per the standard ACP RFD process described in `CONTRIBUTING.md` and `docs/rfds/about.mdx`.

## Why this RFD

`ClientCapabilities` currently exposes only `fs` and `terminal`. We've hit a real downstream gap: some agents gate backend tools by default in isolated stdio mode, with no spec-blessed way for a client to opt in. The RFD body documents the concrete reproducer — Grok Build's interactive (non-ACP) sessions invoke X-search / video-generation backend tools routinely (27 X-search calls in one user's session history), but the same model under `grok agent stdio` cannot reach them. A vendor-specific `_meta.x.ai/requestedBackendTools` workaround was silently ignored by Grok agent 0.2.3.

The captured negative-evidence trace lives at https://github.com/sleep2agi/agent-network/blob/main/docs/rfcs/RFC-021-acp-capability-profile-expansion.md §11 — including the LLM's verbatim self-report ("当前会话中没有连接任何 X 搜索相关的 MCP 工具 ... 通过多次 search_tool 查询均未发现") and the dual-position `_meta` hint that didn't move the needle.

`_meta` alone doesn't generalise — each vendor would need its own key, and even when one agent honours one, no other agent will. The RFD proposes a closed-but-`#[non_exhaustive]` enum of categories (`Search` / `SocialSearch` / `MediaGen` / `Web` / `CodeExec` / `Voice` / ...) so cross-vendor semantics stay anchored.

## Backward compatibility

The field is optional. Agents that don't understand it (current behaviour) ignore it without issue. Clients **MUST NOT** assume any tool is available just because they requested a category — it's purely a hint, never a guarantee.

## Implementation plan (from the RFD)

1. **Phase 1 — this PR**: RFD-only dialogue.
2. **Phase 2**: feature-gated landing behind `unstable_requested_tool_categories` in both `src/v1/client.rs` and `src/v2/client.rs`, JSON schema regenerated, serde round-trip + absent/empty/unknown-variant tests.
3. **Phase 3**: reference honouring in at least one ACP agent (proves the design end-to-end).
4. **Phase 4**: stabilisation — flag removed when ≥2 agents and ≥1 client honour the field in shipped releases.

## FAQ covered in the RFD body

- Why a closed enum vs free-form strings
- Why `_meta` alone doesn't generalise (with the Grok probe as exhibit A)
- The "doesn't this push agents toward exposing what they hide" concern (it's a hint, not a guarantee)
- Three alternatives considered + rejected (`_meta`-only / mandatory exposure / per-tool allowlist)

## Author context

Filing on behalf of the [sleep2agi/agent-network](https://github.com/sleep2agi/agent-network) project — a multi-agent orchestration platform built on ACP. Happy to iterate on the RFD shape, naming, or scope based on feedback.

### DIFF
--- a/docs/rfds/requested-tool-categories.mdx
+++ b/docs/rfds/requested-tool-categories.mdx
@@ -1,0 +1,128 @@
+---
+title: "Requested Tool Categories"
+---
+
+Author(s): [vansin](https://github.com/vansin) (filing on behalf of the [sleep2agi/agent-network](https://github.com/sleep2agi/agent-network) project — multi-agent orchestration platform built on ACP)
+
+## Elevator pitch
+
+> What are you proposing to change?
+
+Add an optional `requested_tool_categories: Option<Vec<RequestedToolCategory>>` field to `ClientCapabilities` so a client can **hint to the agent** which **abstract categories of in-LLM tools** it would like the agent to expose during a session. Categories are vendor-neutral (e.g. `search`, `social_search`, `media_gen`, `web`, `code_exec`, `voice`), and each agent implementation maps a category to whatever concrete tools (built-in or vendor-private) it can offer.
+
+Backward-compatible: the field is optional; agents that don't understand it can ignore it without behaviour change.
+
+## Status quo
+
+> How do things work today and what problems does this cause? Why would we change things?
+
+`ClientCapabilities` currently has only `fs` and `terminal` knobs. There is **no first-class way for a client to express "I want the LLM I'm talking through to have access to backend tools X, Y, Z"**, and `AgentCapabilities` correspondingly has no `availableTools` enumeration the client can read.
+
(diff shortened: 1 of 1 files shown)


benbrandt:
I think given some other items we need to tackle, I think I will close. I don't know how well various agents will handle this already, though I get the desire

PR #1376, opened 2026-06-05
docs(rfd): add v1/v2 durability and reliability expectations to remote transport RFD

alexhancock:
After two transport working group meetings yesterday on the topic of durability & reliability guarantees of the HTTP/WS transport, I propose the following levels of support in both v1 and v2 of ACP


benbrandt (approved):
Thanks for hosting and consolidating! this makes sense. 

One thing I think we need to explore is if message ids + tool calls ids are enough, or if we need something more... but implementations will tell us I think

PR #1489, opened 2026-06-23
Add workspace trust RFD

Nemtecl:
## Summary

- Proposes a standard workspace trust state for session lifecycle responses.
- Adds a `trust/apply` method for Clients to apply a user-selected trust decision.
- Defines trust decisions for current folder, repository, session-only trust, and decline.

## Context

We implemented this workflow in Vibe using ACP extension methods as a proof point. We would love to have this become part of ACP so Clients and Agents can interoperate on trusted-folder UX without custom protocol surfaces.

## Validation

- `npx prettier --check docs/rfds/workspace-trust.mdx`

### DIFF
--- a/docs/rfds/workspace-trust.mdx
+++ b/docs/rfds/workspace-trust.mdx
@@ -1,0 +1,346 @@
+---
+title: "Workspace Trust for Agent-Loaded Project Context"
+---
+
+- Author(s): [@nemtecl](https://github.com/nemtecl)
+
+## Elevator pitch
+
+> What are you proposing to change?
+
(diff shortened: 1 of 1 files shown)


benbrandt (commented):
@Nemtecl in v2, where we allow agents to do more before a prompt, do you think this could fit well into the request permission flow?
I also changed request permission to have a "subject" that we can extend to other things, including this.

My thought being it seems nice if the agent can ask if necessary, rather than the client having to pass it through every time? And the agent could ask for it at other points in the session?

Just a thought, curious what you think

PR #1494, opened 2026-06-23
feat: add optional systemPrompt field to session/new

wpfleger96:
Adds an optional `systemPrompt: Option<String>` field to `NewSessionRequest` (`session/new`), giving clients a standard way to deliver system-level instructions to an agent at session creation.

The semantics are additive: agents incorporate the content alongside their own built-in instructions rather than replacing them, and an agent that ignores the field behaves exactly as before. The field is optional and backward compatible. Per the field's contract, an agent that cannot incorporate the system prompt MUST return a JSON-RPC error and MUST NOT create the session.

This is the reference implementation of the RFD in #1237.

## Changes
- `src/agent.rs` — `system_prompt: Option<String>` on `NewSessionRequest`, a `system_prompt(...)` builder, and round-trip + absent-omission serialization tests.
- `schema/schema.json`, `schema/schema.unstable.json`, `docs/protocol/schema.mdx`, `docs/protocol/draft/schema.mdx` — regenerated so the published artifacts match the Rust types.
- `docs/protocol/session-setup.mdx`, `docs/protocol/draft/session-setup.mdx` — added `systemPrompt` to the parameter bullet list and JSON-RPC example in the narrative docs.

The field serializes as `systemPrompt` on the wire and is omitted entirely when absent.

Related: #1237 (the RFD this implements) and the consuming goose change [aaif-goose/goose#9971](https://github.com/aaif-goose/goose/pull/9971).

### DIFF
--- a/docs/protocol/draft/schema.mdx
+++ b/docs/protocol/draft/schema.mdx
@@ -1217,6 +1217,20 @@ additional roots are activated for the new session.
 <ResponseField name="mcpServers" type={<a href="#mcpserver">McpServer[]</a>} required>
   List of MCP (Model Context Protocol) servers the agent should connect to.
 </ResponseField>
+<ResponseField name="systemPrompt" type={"string | null"} >
+  Optional system-level instructions the client provides for this session.
+
+When present, the agent MUST incorporate this content into its system prompt
+construction alongside its own built-in instructions. It MUST NOT discard its own
+safety guardrails, identity, or internal configuration in favor of client-provided
+content — the semantics are additive, not a replacement. Agents MUST have their own
+default behavior and MUST NOT require a client system prompt to function. If the agent
+cannot incorporate the system prompt, it MUST return a JSON-RPC error and MUST NOT
+create the session.
+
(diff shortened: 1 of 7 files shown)


jalr4ever:
Hope this work can progress quickly.

wpfleger96:
Closing in favor of the ACP+ approach. The goose-side implementation (aaif-goose/goose#9971) is being updated to read `systemPrompt` from `_meta` instead of as a first-class field, which avoids the need for a fork of the ACP schema crate. The spec proposal itself lives in the RFD at agentclientprotocol/agent-client-protocol#1237 and can be pursued independently of the goose implementation.

benbrandt:
Hey all yeah I think the _meta is the right approach for now. I'm sorry, we are focusing a lot of bandwidth on the v2 spec and we already had a lot of drafts that were sitting there without much progress so I have held off from adopting more until we got through the backlog some more. I think we're almost to the point where we can add some more though, and this feature certainly makes sense! Along wiht some other things in session/new

PR #1575, opened 2026-07-02
feat(unstable): preserve unknown elicitation schemas

benbrandt:


benbrandt:
Given json schema this one doesn't make sense...

PR #1662, opened 2026-07-12
Fix/validate url feature gate

marekdkropiewnicki-dotcom:


benbrandt:
I'm not sure what exactly the motivation of this large change is... and there are a bunch of conflicts. Going to close. If you can explain what this is trying to solve we can revisti

PR #1681, opened 2026-07-14
docs: point TypeScript library page at the fluent agent()/client() APIs

chazcb:
## Why

The [TypeScript library page](https://agentclientprotocol.com/libraries/typescript) still directs readers to `AgentSideConnection` and `ClientSideConnection`, both of which are now marked `@deprecated` in the TypeScript SDK. New integrations should start from the fluent `agent()` / `client()` builder APIs instead.

## What changed

- Points readers at the fluent `agent()` / `client()` entry points, linking to the SDK's [runnable examples](https://github.com/agentclientprotocol/typescript-sdk/tree/main/src/examples) and typedoc rather than inlining code examples here — this repo can't typecheck TypeScript snippets, so keeping the page pointer-level (like the Rust/Python/Kotlin/Java library pages) avoids examples drifting out of sync with the SDK
- Adds a "Deprecated classes" section noting that `AgentSideConnection` / `ClientSideConnection` remain for backwards compatibility but new code should use the fluent API

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/docs/libraries/typescript.mdx
+++ b/docs/libraries/typescript.mdx
@@ -19,10 +19,8 @@ or
 [`client()`](https://agentclientprotocol.github.io/typescript-sdk/functions/client.html)
 function to register typed handlers for the ACP methods you support and connect
 to your counterpart. The
-[library README](https://github.com/agentclientprotocol/typescript-sdk#readme)
-walks through both sides, and the
 [examples directory](https://github.com/agentclientprotocol/typescript-sdk/tree/main/src/examples)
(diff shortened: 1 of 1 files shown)


benbrandt (approved):
Thank you!

chazcb:
@benbrandt sorry one more change: I removed the direct examples and pointed to the library, idea is to avoid drift so the ts library remains source of truth. Happy to revert if you'd prefer the examples inline _and_ the links

PR #1702, opened 2026-07-16
docs(rfd): Clarify session id behavior in forking

Jeremy-Michelson:
Edited the expected response from a 'session/fork' request to include the sessionId of the new fork.

I am suggesting this change because I think the client will need the `sessionId` of the result of forking in order to use the new session that was created via the `session/fork` request.


benbrandt (approved):
This is already the case, but fine clarifying the language here

PR #1858, opened 2026-08-03
docs: clarify the session/update variant set is not exhaustive

### LINKED ISSUE #1694: Docs: prompt-turn page lists 5 SessionUpdate variants, schema v1 defines 11
## What happened

I built an ACP client for Grok Build (`grok agent stdio`) working from the
prose docs at agentclientprotocol.com. The prompt-turn page presents these
`sessionUpdate` values for `session/update` notifications:

`plan`, `agent_message_chunk`, `tool_call`, `tool_call_update`, `usage_update`

I treated that list as exhaustive. The first real session immediately
delivered two more:

- `user_message_chunk` (echo of the submitted prompt, with `_meta: {modelId, promptIndex}`)
- `available_commands_update` (fired twice after `session/new`)

I initially logged these as vendor deviations. They are not: both are
defined in `schema/v1/schema.json`, which has 11 variants in the
`SessionUpdate` union:

`user_message_chunk`, `agent_message_chunk`, `agent_thought_chunk`,
`tool_call`, `tool_call_update`, `plan`, `available_commands_update`,
`current_mode_update`, `config_option_update`, `session_info_update`,
`usage_update`

So a client author following the prose page implements 5 of 11 and then
misdiagnoses spec-conformant traffic as off-spec behavior. Captured
traffic and notes: https://github.com/azuriru3/grok-build-vscode/blob/main/docs/ACP-NOTES.md

## Suggested fix

1. List all schema-defined variants on the prompt-turn page, or add a
   pointer that the page shows a common subset and the schema is the
   authority.
2. Add one sentence advising clients to treat the variant set as open for
   forward compatibility. The extensibility page covers unknown methods and
   notifications, but not unknown variants inside `session/update`, and
   that is where a closed switch bites in practice.

## Secondary question, happy to split into its own issue

What is the intended way for an agent to report an upstream provider
failure mid-turn? `StopReason` has no failure value, which suggests a
JSON-RPC error on `session/prompt`. That is what Grok Build does:

```json
{"jsonrpc":"2.0","id":3,"error":{"code":-32603,"message":"Internal error","data":{"message":"API error (status 403 Forbidden): permission-denied: ...","http_status":403}}}
```

If a JSON-RPC error is the intended mechanism, a sentence saying so in the
prompt-turn page's stop-reason section would remove the ambiguity.

somtri:
Fixes #1694.

The prompt-turn page walks a turn through five `session/update` variants, while `schema/v1/schema.json` defines eleven. Nothing marks the walkthrough as partial, so it reads as the full set.

Adds a Note pointing at the `SessionUpdate` schema reference, plus a short section in `extensibility.mdx` on unrecognized variants, which that page didn't cover. The `draft/` copies get the same change, matching 8750b93.

Two calls worth flagging: I kept the extensibility wording advisory rather than **SHOULD**, since `SessionUpdate` is a closed `oneOf` with no catch-all. And v2 has the same gap (9 shown against 17) which I left alone, happy to add it here or in a follow-up.

The StopReason question in the issue is a separate protocol matter, untouched here.

### DIFF
--- a/docs/protocol/v1/draft/extensibility.mdx
+++ b/docs/protocol/v1/draft/extensibility.mdx
@@ -132,3 +132,7 @@ Implementations **SHOULD** use the `_meta` field in capability objects to advert
 ```
 
 This allows implementations to negotiate custom features during initialization without breaking compatibility with standard Clients and Agents.
+
+## Unrecognized Session Update Variants
+
+Every `session/update` notification carries a `sessionUpdate` field that identifies the variant. Future protocol versions may define new variants. Clients are encouraged to ignore variants they do not recognize rather than treating them as errors. The [`SessionUpdate`](/protocol/v1/draft/schema#sessionupdate) schema reference lists the variants defined in this version.
--- a/docs/protocol/v1/draft/prompt-turn.mdx
+++ b/docs/protocol/v1/draft/prompt-turn.mdx
@@ -105,6 +105,13 @@ Upon receiving the prompt request, the Agent processes the user's message and se
 
 The Agent reports the model's output to the Client via `session/update` notifications. This may include the Agent's plan for accomplishing the task:
 
+<Note>
+  This page walks through the `session/update` variants most common in a prompt
+  turn. It is not the full set. See the
+  [`SessionUpdate`](/protocol/v1/draft/schema#sessionupdate) schema reference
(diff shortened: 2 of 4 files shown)


benbrandt:
Superseded by #2214

PR #1891, opened 2026-08-09
docs(unstable-v2): list all SessionUpdate variants on the prompt lifecycle page

tarikermis:
Refs #1694 and complements #1858, which fixed the v1 page.

## What changed

The v2 prompt-lifecycle pages show JSON examples for the common `session/update` variants, but they never say the walkthrough is incomplete. The schemas currently define 16 named variants plus `other` in stable v2, and 17 plus `other` in the unstable schema.

This updates both pages to match their schemas:

- `docs/protocol/v2/prompt-lifecycle.mdx`: all 16 stable variants.
- `docs/protocol/v2/draft/prompt-lifecycle.mdx`: all 17 unstable variants, including `plan_removed` and the unstable `usage` field on idle state updates.

Each page now has a note explaining that the walkthrough covers common prompt-turn updates, a complete variant table with key fields and links to the relevant topic pages, and a short explanation of custom `_`-prefixed variants and the preserve-and-ignore rule.

## Verification

- Extracted both variant sets from the schema JSON and checked the tables against them: 16/16 stable and 17/17 unstable, with no extras.
- Cross-checked the key fields against the variant definitions.
- Verified every internal link and anchor in both documentation trees.
- Prettier passes on both files and the repository spellcheck passes.

I did not run `mint dev`; the table and `<Note>` syntax already exist on these pages and were checked statically.

Best Regards, Tarik

### DIFF
--- a/docs/protocol/v2/draft/prompt-lifecycle.mdx
+++ b/docs/protocol/v2/draft/prompt-lifecycle.mdx
@@ -158,6 +158,13 @@ When foreground work starts or resumes, the Agent **MUST** send a `state_update`
 
 The language model **MAY** respond with text content, tool calls, or both.
 
+<Note>
+  This walkthrough covers the `session/update` variants most common in a prompt
+  turn. It is not the full set: [Session Update
+  Variants](#session-update-variants) lists every variant the schema defines in
+  this version.
+</Note>
+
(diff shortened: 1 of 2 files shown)


benbrandt:
Superseded by #2214

PR #1977, opened 2026-08-18
Update providers RFD to support batch update and restore

Ololoshechkin:
Checked with how junie, codex and claude implement this and work with providers, and it comes out that individual updates are not required, and providerID concept doesn't seem to represent the reality of how they connect to providers: instead, the most natural model is to expose protocols/APIs that agent supports and either update them or restore to default settings.


agent-client-protocol-schema/src/v1/agent.rs
```diff
@@ -3325,8 +3325,8 @@ pub enum LlmProtocol {
 #[serde(rename_all = "camelCase")]
 #[non_exhaustive]
 pub struct ProviderCurrentConfig {
-    /// Protocol currently used by this provider.
-    pub api_type: LlmProtocol,
+    /// Name of the provider to be shown in UI.
+    pub presentable_name: String,
```
benbrandt, line 3329: Can this just be `name`? To be more inline with other entities in ACP?

agent-client-protocol-schema/src/v1/agent.rs
```diff
@@ -3400,15 +3378,8 @@ impl ProviderId {
 #[serde(rename_all = "camelCase")]
 #[non_exhaustive]
 pub struct ProviderInfo {
-    /// Provider identifier, for example "main" or "openai".
-    pub provider_id: ProviderId,
```
benbrandt, line 3404: Why did we remove id? It seems necessary to have in case you have two openai compatible endpoints? (I believe we went over this in previous discussions?)

I think having an id has proved very important time and again in other parts of the protocol, and I'm kind of worried about losing it here.

docs/rfds/custom-llm-endpoint.mdx
```diff
@@ -2,17 +2,17 @@
...
 
 > What are you proposing to change?
 
-Add the ability for clients to discover and configure agent LLM providers (identified by `id`) via dedicated provider methods:
+Add the ability for clients to discover and configure agent LLM providers (identified by API type) via dedicated provider methods:
 
-- `providers/list`
-- `providers/set`
-- `providers/disable`
+- `providers/list_api`
```
benbrandt, line 13: Nitpick, but why not just leave it as "list"? Or is this supposed to rather just return the options available, not what is currently configured?

benbrandt (commented):
How does this interact with the auth flow? That part is still kind of missing for me. Especially if we do things like auth/status or something where you ask if it is authenticated.

PR #1992, opened 2026-08-20
Add subagents RFD

Ololoshechkin:
Add subagents RFD

### DIFF
--- a/agent-client-protocol-schema/src/v1/client.rs
+++ b/agent-client-protocol-schema/src/v1/client.rs
@@ -585,6 +585,10 @@ impl SubagentSessionCapabilities {
 /// This capability is not part of the spec yet, and may be removed or changed at any point.
 ///
 /// Notification that a subagent reached a terminal state.
+///
+/// Sent on the immediate parent session after all child session updates and
+/// after every pending permission or elicitation request issued for the child
+/// has resolved.
 #[cfg(feature = "unstable_subagents")]
 #[serde_as]
 #[skip_serializing_none]
--- a/agent-client-protocol-schema/src/v1/client.rs
+++ b/agent-client-protocol-schema/src/v1/client.rs
@@ -437,6 +437,211 @@ impl CompactionSummaryChunk {
         self.meta = meta.into_option();
         self
     }
+    /// **UNSTABLE**
+    ///
+    /// This capability is not part of the spec yet, and may be removed or changed at any point.
+    ///
+    /// Notification that the session spawned a subagent.
+    #[cfg(feature = "unstable_subagents")]
+    SubagentSpawned(SubagentSpawnedUpdate),
+    /// **UNSTABLE**
+    ///
+    /// This capability is not part of the spec yet, and may be removed or changed at any point.
+    ///
+    /// Notification that a subagent reached a terminal state.
+    #[cfg(feature = "unstable_subagents")]
+    SubagentStateUpdate(SubagentStateUpdate),
+}
+
+/// **UNSTABLE**
+///
+/// This capability is not part of the spec yet, and may be removed or changed at any point.
+///
+/// Notification that a session spawned a subagent.
+#[cfg(feature = "unstable_subagents")]
+#[serde_as]
+#[skip_serializing_none]
+#[cfg_attr(feature = "schemars", derive(schemars::JsonSchema))]
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq)]
+#[serde(rename_all = "camelCase")]
+#[non_exhaustive]
+pub struct SubagentSpawnedUpdate {
+    /// The opaque session ID used by all subsequent updates for the child.
+    pub subagent_session_id: SessionId,
+    /// A short, human-readable label for the subagent.
+    pub name: String,
+    /// A human-readable summary of the work delegated to the subagent.
+    pub task: String,
+    /// Client-to-agent operations permitted for this subagent session.
+    pub capabilities: SubagentSessionCapabilities,
+    /// The _meta property is reserved by ACP to allow clients and agents to attach additional
+    /// metadata to their interactions. Implementations MUST NOT make assumptions about values at
+    /// these keys.
+    ///
+    /// See protocol docs: [Extensibility](https://agentclientprotocol.com/protocol/extensibility)
+    #[serde_as(deserialize_as = "DefaultOnError")]
+    #[cfg_attr(feature = "schemars", schemars(extend("x-deserialize-default-on-error" = true)))]
+    #[serde(default)]
+    #[serde(rename = "_meta")]
+    pub meta: Option<Meta>,
+}
+
+#[cfg(feature = "unstable_subagents")]
+impl SubagentSpawnedUpdate {
+    /// Builds a subagent spawn update with all required fields set.
+    #[must_use]
+    pub fn new(
+        subagent_session_id: impl Into<SessionId>,
+        name: impl Into<String>,
+        task: impl Into<String>,
+        capabilities: SubagentSessionCapabilities,
+    ) -> Self {
+        Self {
+            subagent_session_id: subagent_session_id.into(),
+            name: name.into(),
+            task: task.into(),
+            capabilities,
+            meta: None,
+        }
+    }
+
+    /// The _meta property is reserved by ACP to allow clients and agents to attach additional
+    /// metadata to their interactions. Implementations MUST NOT make assumptions about values at
+    /// these keys.
+    #[must_use]
+    pub fn meta(mut self, meta: impl IntoOption<Meta>) -> Self {
+        self.meta = meta.into_option();
+        self
+    }
+}
+
+/// **UNSTABLE**
+///
+/// This capability is not part of the spec yet, and may be removed or changed at any point.
+///
+/// Client-to-agent operations permitted for a specific subagent session.
+#[cfg(feature = "unstable_subagents")]
+#[serde_as]
+#[skip_serializing_none]
+#[cfg_attr(feature = "schemars", derive(schemars::JsonSchema))]
+#[derive(Default, Debug, Clone, Serialize, Deserialize, PartialEq, Eq)]
+#[serde(rename_all = "camelCase")]
+#[non_exhaustive]
+pub struct SubagentSessionCapabilities {
+    /// Whether the client may cancel this subagent. Omission is equivalent to `false`.
+    #[serde_as(deserialize_as = "DefaultOnError")]
+    #[cfg_attr(feature = "schemars", schemars(extend("x-deserialize-default-on-error" = true)))]
+    #[serde(default, skip_serializing_if = "std::ops::Not::not")]
+    pub cancel: bool,
+    /// Whether the client may close this subagent. Omission is equivalent to `false`.
+    #[serde_as(deserialize_as = "DefaultOnError")]
+    #[cfg_attr(feature = "schemars", schemars(extend("x-deserialize-default-on-error" = true)))]
+    #[serde(default, skip_serializing_if = "std::ops::Not::not")]
+    pub close: bool,
+    /// The _meta property is reserved by ACP to allow clients and agents to attach additional
+    /// metadata to their interactions. Implementations MUST NOT make assumptions about values at
+    /// these keys.
+    #[serde_as(deserialize_as = "DefaultOnError")]
+    #[cfg_attr(feature = "schemars", schemars(extend("x-deserialize-default-on-error" = true)))]
+    #[serde(default)]
+    #[serde(rename = "_meta")]
+    pub meta: Option<Meta>,
+}
+
+#[cfg(feature = "unstable_subagents")]
+impl SubagentSessionCapabilities {
+    /// Builds an empty capability set; cancellation and close are disabled.
+    #[must_use]
+    pub fn new() -> Self {
+        Self::default()
+    }
+
+    /// Whether the client may cancel this subagent.
+    #[must_use]
+    pub fn cancel(mut self, cancel: bool) -> Self {
+        self.cancel = cancel;
+        self
+    }
+
+    /// Whether the client may close this subagent.
+    #[must_use]
+    pub fn close(mut self, close: bool) -> Self {
+        self.close = close;
+        self
+    }
+
+    /// The _meta property is reserved by ACP to allow clients and agents to attach additional
+    /// metadata to their interactions. Implementations MUST NOT make assumptions about values at
+    /// these keys.
+    #[must_use]
+    pub fn meta(mut self, meta: impl IntoOption<Meta>) -> Self {
+        self.meta = meta.into_option();
+        self
+    }
+}
+
+/// **UNSTABLE**
+///
+/// This capability is not part of the spec yet, and may be removed or changed at any point.
+///
+/// Notification that a subagent reached a terminal state.
+#[cfg(feature = "unstable_subagents")]
+#[serde_as]
+#[skip_serializing_none]
+#[cfg_attr(feature = "schemars", derive(schemars::JsonSchema))]
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq)]
+#[serde(rename_all = "camelCase")]
+#[non_exhaustive]
+pub struct SubagentStateUpdate {
+    /// The session ID of the subagent whose state changed.
+    pub subagent_session_id: SessionId,
+    /// The terminal state reached by the subagent.
+    pub state: SubagentState,
+    /// The _meta property is reserved by ACP to allow clients and agents to attach additional
+    /// metadata to their interactions. Implementations MUST NOT make assumptions about values at
+    /// these keys.
+    #[serde_as(deserialize_as = "DefaultOnError")]
+    #[cfg_attr(feature = "schemars", schemars(extend("x-deserialize-default-on-error" = true)))]
+    #[serde(default)]
+    #[serde(rename = "_meta")]
+    pub meta: Option<Meta>,
+}
+
+#[cfg(feature = "unstable_subagents")]
+impl SubagentStateUpdate {
+    /// Builds a terminal state update with all required fields set.
+    #[must_use]
+    pub fn new(subagent_session_id: impl Into<SessionId>, state: SubagentState) -> Self {
+        Self {
+            subagent_session_id: subagent_session_id.into(),
+            state,
+            meta: None,
+        }
+    }
+
+    /// The _meta property is reserved by ACP to allow clients and agents to attach additional
+    /// metadata to their interactions. Implementations MUST NOT make assumptions about values at
+    /// these keys.
+    #[must_use]
+    pub fn meta(mut self, meta: impl IntoOption<Meta>) -> Self {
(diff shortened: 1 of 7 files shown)


nikita-ashihmin:
Implementation feedback from AIR and the reference adapter branches:

- https://github.com/agentclientprotocol/claude-agent-acp/pull/1017
- https://github.com/agentclientprotocol/codex-acp/pull/419

The canonical capability and lifecycle shape works for both adapters. We only needed `_meta.jetbrains.air.capabilities = ["nativeSubagentSessions"]` as a temporary negotiation bridge because the released SDKs currently remove the draft `subagents` fields. We treat this metadata as a compatibility bridge, not as an alternative protocol. It may help to make SDK preservation of the draft fields an explicit rollout prerequisite.

Two edge cases still seem underspecified:

1. The RFD defines orphan recovery during `session/load`, but it does not define the immediate Client state when a live connection ends or `session/prompt` returns without the required terminal update. Clients can otherwise choose failed, keep running forever, or show disconnected. Could the RFD require a local disconnected state for an unknown outcome and clarify whether the Client emits no wire update?

2. A child can own pending `session/request_permission` or `session/create_elicitation` requests when the child or its parent is cancelled. Could the RFD require those requests to resolve or cancel before `subagent_state_update`, and define how a Client handles a response that races with the terminal update? This prevents a terminal child from retaining a live control or moving that control to the wrong history.

One smaller compatibility note: the current RFD requires `task`. AIR now treats the older `description` field only as a temporary input fallback. New adapters emit `task`.

Ololoshechkin:
@nikita-ashihmin please check again :)

benbrandt:
I would like to keep the RFD + protocol changes in separate PRs because it is easier to accept an RFD without having to do rounds on the Rust code at the same time to get a merge

agent-client-protocol-schema/src/v1/client.rs
```diff
@@ -439,6 +453,203 @@ impl CompactionSummaryChunk {
...
+#[non_exhaustive]
+pub struct SubagentSpawnedUpdate {
+    /// The opaque session ID used by all subsequent updates for the child.
+    pub subagent_session_id: SessionId,
+    /// A short, human-readable label for the subagent.
+    pub name: String,
+    /// A human-readable summary of the work delegated to the subagent.
+    pub task: String,
+    /// Client-to-agent operations permitted for this subagent session.
+    pub capabilities: SubagentSessionCapabilities,
```
benbrandt, line 476: I would imaging several of these should be optional

agent-client-protocol-schema/src/v1/client.rs
```diff
@@ -439,6 +453,203 @@ impl CompactionSummaryChunk {
...
+    }
+}
+
+/// Terminal state of an announced subagent.
+#[cfg(feature = "unstable_subagents")]
+#[cfg_attr(feature = "schemars", derive(schemars::JsonSchema))]
+#[derive(Debug, Clone, Copy, Serialize, Deserialize, PartialEq, Eq, Hash)]
+#[serde(rename_all = "snake_case")]
+#[non_exhaustive]
+pub enum SubagentState {
```
benbrandt, line 642: My inclination is we should treat this similarly to how we are treating tool calls, messages, and terminals in v2 where we continuously update a given subagent "entity" so we only have one notification type with clear upsert/update semantics

docs/rfds/subagents.mdx
```diff
@@ -0,0 +1,487 @@
...
+
+Add an optional `subagents` object to `AgentCapabilities.sessionCapabilities`.
+Its presence means the Agent may expose agent-created subagents. The field is
+optional and non-nullable. An omitted field means the capability is not
+supported; `{}` means it is supported.
+
+```json
+{
+  "clientCapabilities": {
+    "subagents": {}
```
benbrandt, line 66: I think this is good for v1, but just a note, I think for v2 we can assume support

docs/rfds/subagents.mdx
```diff
@@ -0,0 +1,487 @@
...
+> What are you proposing to improve the situation?
+
+### Capability negotiation
+
+Add an optional `subagents` object to `ClientCapabilities`. Its presence means
+the Client understands the session updates and restricted-session semantics in
+this RFD. The field is optional and non-nullable. An omitted field means the
+capability is not supported; `{}` means it is supported.
+
+Add an optional `subagents` object to `AgentCapabilities.sessionCapabilities`.
```
benbrandt, line 58: Why does the agent need to express support? Wouldn't the notifications inherently handle it?

docs/rfds/subagents.mdx
```diff
@@ -0,0 +1,487 @@
...
+An Agent **MUST NOT** send the updates defined by this RFD unless both parties
+advertised `subagents`. It may still use subagents internally and present their
+results through the parent session.
+
+### Announcing a subagent
+
+When an Agent creates a subagent that it wants to expose, it **MUST** send a
+`subagent_spawned` update on the immediate parent session before sending any
+updates for the child session:
+
```
benbrandt, line 90: Yeah this seems important. We may want to handle some things on the SDK side to better tolerate queueing messages for sessions we don't know about in case someone doesn't abide by this, but this does seem also easy enough (before the first event for the new session, emit the first subagent notification)

docs/rfds/subagents.mdx
```diff
@@ -0,0 +1,487 @@
...
+```json
+{
+  "jsonrpc": "2.0",
+  "method": "session/update",
+  "params": {
+    "sessionId": "sess_parent",
+    "update": {
+      "sessionUpdate": "subagent_spawned",
+      "subagentSessionId": "sess_child_1",
```
benbrandt, line 99: I'm torn on if we need subagentSessionId here since it is implicitly the session id of the subagent based on the structure of this udpate... So I feel like this could just be `sessionId`

docs/rfds/subagents.mdx
```diff
@@ -0,0 +1,487 @@
...
+```
+
+`subagentSessionId`, `name`, `task`, and `capabilities` are required and
+non-nullable:
+
+- `subagentSessionId` is an opaque `SessionId` unique within the ACP
+  connection. It identifies the child in all subsequent ACP messages.
+- `name` is a short, human-readable label. It need not be unique.
+- `task` is a human-readable summary of the work delegated to the child. It is
+  descriptive, not a prompt that the Client can edit or resubmit.
```
benbrandt, line 118: These feel optional to me. Like it is nice if we have it, but also doesnt' feel necessary at the protocol level

docs/rfds/subagents.mdx
```diff
@@ -0,0 +1,487 @@
...
+- `capabilities` describes the Client-to-Agent operations permitted for this
+  specific child session. `cancel` and `close` are optional, non-nullable
+  booleans whose omission is equivalent to `false`.
```
benbrandt, line 121: These are definitely nice hints to avoid errors. However, I feel like we may want a few more than this...

Like `prompt` or whether you can change config options etc... But we can also add later.

Should we assume the agent will take care of closing the subagents though? Why would the user close just the subagent if the parent agent is managing the lifecycle?

Just curious to know more I guess about why you started with these two and which use cases you are accounting for here?

docs/rfds/subagents.mdx
```diff
@@ -0,0 +1,487 @@
...
+
+The outer `sessionId` establishes the immediate parent. This supports arbitrary
+nesting without adding a second parent identifier. If a child spawns another
+subagent, the Agent sends `subagent_spawned` with the child's session ID as the
+outer `sessionId`.
```
benbrandt, line 126: Yep this seems right

docs/rfds/subagents.mdx
```diff
@@ -0,0 +1,487 @@
...
+
+The Agent **MUST** send the spawn update even if it expects the child to finish
+very quickly. The Client **MUST** create the child session before processing
```
benbrandt, line 129: I wonder how much this Client "must" is necessary... I think the client can handle these updates how it likes as long as it can guarantee agent behavior.

docs/rfds/subagents.mdx
```diff
@@ -0,0 +1,487 @@
...
+later updates bearing that session ID.
+
+The child inherits the parent's effective working directory, additional
+directories, MCP servers, and Client capabilities. The Agent may apply stricter
+runtime or tool restrictions to the child, but it **MUST NOT** grant access that
+was not available to the parent. A future RFD may add explicit per-child
+execution-context fields if Clients need to display or negotiate them.
```
benbrandt, line 136: Hmm I don't know if we want to do this? The reason I say this is I see codex spawning lots of subagents in /tmp directories to do work, and may not pass along mcp or client capabilities.

We likely need a way to express these though

docs/rfds/subagents.mdx
```diff
@@ -0,0 +1,487 @@
...
+sending a `subagent_state_update` on its immediate parent session:
+
+```json
+{
+  "jsonrpc": "2.0",
+  "method": "session/update",
+  "params": {
+    "sessionId": "sess_parent",
+    "update": {
+      "sessionUpdate": "subagent_state_update",
```
benbrandt, line 187: Again, I wonder if we want a more generic `subagent_update` notification that captures all updates of that subagent (see v2 patterns that we can backport to v1 here)

docs/rfds/subagents.mdx
```diff
@@ -0,0 +1,487 @@
...
+- the parent `session/prompt` request returns — with any stop reason or with an
+  error — before the child's terminal update was received. Under the
+  [v1 scope rule](#scope-in-acp-v1) this is an Agent bug, but Clients still
+  need a defined outcome.
+
+In both cases the Client **MUST** immediately place the child and its announced
+descendants in a local `disconnected` state: the task outcome is unknown. The
+Client **MUST NOT** present such a child as `failed` or `cancelled`, **MUST
+NOT** continue presenting it as running, and **MUST NOT** wait indefinitely for
+a terminal update.
```
benbrandt, line 227: Hmm this seems odd to me... Because the client could continue to receive updates from that session, which would tell it whether it is still running or not?

docs/rfds/subagents.mdx
```diff
@@ -0,0 +1,487 @@
...
+that subagent and its descendants. It does not cancel its parent or siblings.
+After cancellation finishes, the Agent sends `subagent_state_update` with
+`state: "cancelled"`.
+
+Unlike cancellation of a user-facing session, there is no child
+`session/prompt` request to complete with a `cancelled` stop reason. The
+terminal `subagent_state_update` is the Client's confirmation that cancellation
+finished.
+
+If `capabilities.close` is `true`, and the Agent also advertised the existing
```
benbrandt, line 297: Cancellation I can understand in case you want to stop some task, but I am still confused when you would want to close it?

agent-client-protocol-schema/src/v1/client.rs
```diff
@@ -439,6 +453,203 @@ impl CompactionSummaryChunk {
...
+    /// The subagent completed its task successfully.
+    Completed,
+    /// The subagent failed to complete its task.
+    Failed,
+    /// The subagent was cancelled.
+    Cancelled,
+    /// The Agent lost the child runtime and cannot determine its task outcome.
+    Disconnected,
```
benbrandt, line 650: Another question: how much do we need this state if we already have the general state of the session and the prompt in question?

more important for cases where there parent agent is able to do multi-turn interactions (quite common at this point) and reprompt on its own.

benbrandt (commented):
I think this is going in the right direction overall, but still have several questions I think we need to resolve. 

Thanks for kicking this off!

docs/rfds/subagents.mdx
Ololoshechkin: Why do you think so? Not all possible agents even support subagents (and not all of them plan/should do so afaik)

Ololoshechkin:
Thanks for the thorough review, Ben! Pushed a rework addressing all threads (182be75, ee65f9c, a06ecf2):

- `subagent_spawned` + `subagent_state_update` are now a single upsert-style `subagent_update` following the v2 entity pattern (only `subagentSessionId` required; explicit `running` state)
- Agent-side capability and per-child `close` are gone
- Execution context is no longer implied to be inherited
- SDKs may buffer updates for unannounced children
- The v2 schema now carries the same type: capability-free, v2 patch semantics (`null` clears), open state enum — plus a new “Subagents in ACP v2” section in the RFD

On splitting RFD and protocol changes: happy to do that — I'd keep this PR as the RFD and move the schema/Rust changes into a follow-up PR once you're happy with the shape. Say the word and I'll split it.

@nikita-ashihmin heads-up: the wire shape changed (single `subagent_update`, optional fields, no per-child `close`), so the adapter branches will need a small migration. Your three earlier points survived the rework: SDK preservation is still an explicit rollout prerequisite, the local `disconnected` rule and the pending-request race handling are unchanged (now with added rationale), and `task` remains the canonical field — but note it is now optional, so fallback code shouldn't assume it is always present.

agent-client-protocol-schema/src/v1/client.rs
Ololoshechkin: Done in ee65f9c — only `subagentSessionId` is required now. `name`/`task`/`capabilities`/`state` are optional with patch semantics: omitted (or `null`) means unchanged, and never-supplied has defined defaults (client-chosen label, no permitted operations, `running`).

agent-client-protocol-schema/src/v1/client.rs
Ololoshechkin: Agreed — merged `subagent_spawned` + `subagent_state_update` into a single upsert-style `subagent_update` modeled on v2's `TerminalUpdate`/`ToolCallUpdate`: only the ID is required, the first update announces the child, later ones patch it, and one of them carries the terminal state (ee65f9c). Also mirrored the same type into the v2 schema — capability-free, `MaybeUndefined` patch semantics, open state enum — so the shape carries into v2 unchanged (a06ecf2).

docs/rfds/subagents.mdx
Ololoshechkin: Resolving my own question above after implementing it: “assume support” turned out fine — it doesn't force Agents to have subagents; an Agent without them simply never sends the update. What v2 removes is only the negotiation handshake: unknown `sessionUpdate` types flow into `OtherSessionUpdate`, which Clients preserve and otherwise ignore. The RFD now has a “Subagents in ACP v2” section, and the v2 schema carries `subagent_update` behind the same unstable flag (a06ecf2).

docs/rfds/subagents.mdx
Ololoshechkin: You're right — dropped the Agent-side capability entirely (182be75, ee65f9c). Subagent traffic only flows Agent → Client, so the Client capability alone prevents sending updates the other side can't understand; a Client learns that an Agent exposes subagents by receiving the first update.

docs/rfds/subagents.mdx
Ololoshechkin: Added exactly that: SDKs MAY buffer updates for unknown session IDs until the announcing update arrives, and Agents MUST NOT rely on such tolerance. Also added to the implementation plan.

docs/rfds/subagents.mdx
Ololoshechkin: I kept `subagentSessionId`, though not strongly attached: the update travels on the parent's stream where `params.sessionId` already denotes the parent, and reusing the same key for a different session in one message felt like a routing-bug magnet. Added an FAQ entry with this reasoning — happy to rename if you still prefer `sessionId`.

docs/rfds/subagents.mdx
Ololoshechkin: Made both optional (182be75). If `name` was never supplied the Client chooses its own fallback presentation; `task` is optional and can be revised mid-run under the new patch semantics.

docs/rfds/subagents.mdx
Ololoshechkin: Started from the smallest control the reference adapters could actually validate: stopping a runaway worker, which maps 1:1 onto the existing `session/cancel`. You convinced me on `close` — dropped it (see the other thread). `capabilities` stays an object precisely so `prompt`/`steer`/config controls can be added compatibly once interoperable semantics emerge; added an FAQ recording this.

docs/rfds/subagents.mdx
Ololoshechkin: Kept as is 👍

docs/rfds/subagents.mdx
Ololoshechkin: Agreed — removed the prescriptive Client-side MUST. The Agent-side ordering guarantee is the protocol contract; how a Client tracks or renders announced children is now explicitly its own concern.

docs/rfds/subagents.mdx
Ololoshechkin: Reworked (182be75): the child's execution context — cwd, MCP servers, tools — is now explicitly Agent-chosen and may differ from the parent's (scratch dirs like the Codex `/tmp` case). Kept only the enforceable part: child activity flows through the same connection, subject to the same Client capabilities and permission checks, and a child must not be used to bypass parent-session restrictions. The future-RFD pointer for explicit per-child context fields stays — agreed we'll likely want a way to express these; an optional `workingDirectory` on `subagent_update` would be an easy follow-up if you want it now.

docs/rfds/subagents.mdx
Ololoshechkin: Done — folded both notifications into the single upsert-style `subagent_update` (ee65f9c), mirrored into v2 (a06ecf2).

docs/rfds/subagents.mdx
Ololoshechkin: The constraint driving this: v1 ties `session/update` to the active prompt turn, so once `session/prompt` returns there's no defined channel through which the child's real outcome could still arrive — “keep waiting” becomes a spinner that can never resolve (this rule came from AIR's implementation feedback earlier in this PR). A late terminal update on a still-live connection is already allowed to supersede the local state. I added this rationale to the section. In v2 this trigger disappears entirely: prompt resolves at acceptance and updates keep flowing, so connection loss remains the only cause (covered in the new v2 section).

docs/rfds/subagents.mdx
Ololoshechkin: I couldn't defend it either — removed per-child `close` entirely (182be75, ee65f9c). `session/close` is now in the forbidden list for child session IDs; closing the parent (existing `sessionCapabilities.close`) is how resources get released. The FAQ records the reasoning, and `capabilities` can grow a `close` flag later if a concrete need appears.

agent-client-protocol-schema/src/v1/client.rs
Ololoshechkin: Kept it, with an FAQ making the case: `completed`/`failed`/`cancelled` are claims about the delegated task's outcome, and after a crash or reload an Agent can honestly not know which announced child ended how — the parent's own state (or its ability to re-prompt) can't express that per child. Without `disconnected` the Agent must guess (`failed`), lie (`cancelled`), or leave the child running forever in replay. Open to collapsing it into something else, but every alternative seemed to misrepresent history.

PR #2043, opened 2026-08-26
feat(unstable-v2): add session injection schema

danielkov:
## Summary

Adds unstable ACP v2 schema for sending user messages to a session that is already running.

- `session/inject` accepts content plus `queue` or `steer` delivery mode and returns an agent-owned message ID
- `session/revoke_inject` cancels a pending message by ID
- `session/replace_inject` replaces pending content without changing its ID or queue position
- session capabilities advertise supported modes, stream behavior, revoke, and replace support
- delivery uses the existing v2 `user_message` session update
- typed errors distinguish unknown, delivered, and unsupported operations
- all additions are behind `unstable_session_inject`; stable schema artifacts do not change

## Related

- RFD: https://github.com/agentclientprotocol/agent-client-protocol/pull/1261
- Rust SDK: https://github.com/agentclientprotocol/rust-sdk/pull/339
- AgentKit: https://github.com/danielkov/agentkit/pull/13
- Kit: https://github.com/speakeasy-api/kit/pull/24


mkorun:
Sharing some live, empirical data that might be useful here, since this PR is where these semantics actually land in the schema.

We ran a controlled comparison of four CLI coding agents (Codex CLI, Claude Code, OpenCode, Google's Antigravity CLI) against the same steering scenario — a multi-step task, corrected mid-run with "stop after step N, don't do the rest" — and the results map fairly cleanly onto `SessionInjectSteerInStream`:

- **Codex CLI** (its own `turn/steer` RPC): behaves like `interrupt` — lands near-instantly, mid-generation.
- **Claude Code**: no dedicated steer RPC at all — writing a new user message onto the persistent `--input-format stream-json` stdin while a turn is running gets folded in live, but only at the next safe boundary, closer to `finish` than `interrupt`. This also confirms a concern @SteffenDE raised on the RFD thread (#1261) about the Claude Agent SDK only generating a message ID at delivery time: there's no separate pending-ack step in the CLI's own mechanism at all, it's a raw stdin write — an ACP adapter really would have to synthesize `InjectSessionResponse.messageId` itself rather than getting one back synchronously.
- **OpenCode**: has genuine mid-turn steering through its own session API too, also `finish`-like — worth noting since neither OpenCode nor Antigravity come up anywhere in the RFD discussion.
- **Antigravity CLI**: no live steering mechanism at all, confirmed directly against the CLI, and no roadmap signal either.

For the two without live steering, our fallback measurements might be useful for gauging how costly the plain `queue`-after-cancel path actually is in practice: giving the resumed turn the *actual correction* as its real next prompt (not a generic "continue") correctly handles genuine redirection, not just abort — verified live. It's not free though: it pays for the full prior context again as fresh input, plus real wall-clock restart time (process teardown + reinit) — structural costs, independent of whether the caller times the cancel to land on a clean step boundary.

Happy to share more detail on methodology if it's useful for pinning `interrupt`/`finish` against real agent behavior.

